PR #31, opened 2022-10-12
docs: Add korean i18n support

anaclumos:
## 어떤 목적의 PR인가요?

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Other : 

## 내용을 설명해주세요

Docusaurus의 i18n 기능을 활용해 SEO 및 사용자 친화적인 i18n을 구현합니다.

## 체크리스트

- [x] 아래 항목들을 모두 확인하였습니다.

1. PR 내용을 잘 설명하였습니다
2. 문서화를 충분히 하였습니다 (필요시)
3. 테스트가 충분히 작성되었습니다 (필요시)


anaclumos:
We might want to change index.md(x) into index.jsx/tsx to fully take advantage of docusaurus/translate(). Any thoughts?

raon0211:
We were actually planning to work on i18n -- thanks for reminding us! 

However we were planning to support i18n by managing two separate `.md` files, namely `en.md` and `ko.md` per page. We didn't think about using `<Translate />` components. Which do you think would be a better option?

anaclumos:
Yes, I agree with your decision! I also discussed this matter with the maintainers of Docusaurus and suggested a similar workflow. For reference: https://github.com/facebook/docusaurus/discussions/7608

It seems there are misconceptions about docusaurus/translate. Docusaurus Translate is for translating static labels and strings. As seen in my PR, Docusaurus extracts them into easily editable JSON files. In a way, it is Android's string.xml approach + JS's string interpolation.

For the documents themselves, as you suggested, we can go with the markdown approach. Docusaurus will create the i18n directory, and inside, we can populate localized markdown files.

anaclumos:
Just an FYI, This is my current workflow for handling i18n. https://github.com/anaclumos/extracranial.

For maintenance reasons, I keep both en.md and ko.md inside a source folder (in this case, the 'posts' folder only because I do not translate docs). That way, I can manage both en.md and ko.md inside one working directory.

A python script will populate the files to their appropriate position on build time. One side effect is that the last edited time of docs will mess up (because technically, python script performs file editing on build time, overwriting all records).

raon0211:
Oh, I agree with you in that we should translate static labels with JSON files. 

I was just worried we will write library docs in [this style](https://github.com/toss/slash/pull/31/files#diff-e21e58ba504b9adc10039b3f8bf6939cf003cf21c26ace30807b99bf5c0229e1R9), using `<Translate />` components. I thought it will be hard to write and maintain.

raon0211:
>Just an FYI, This is my current workflow for handling i18n. https://github.com/anaclumos/extracranial.
>
> For maintenance reasons, I keep both en.md and ko.md inside a source folder (in this case, the 'posts' folder only because I do not translate docs). That way, I can manage both en.md and ko.md inside one working directory.
>
>A python script will populate the files to their appropriate position on build time. One side effect is that the last edited time of docs will mess up (because technically, python script performs file editing on build time, overwriting all records).

Yeah, actually I was reading your blog and thought it was awesome (😄). We were planning to manage our documents in that style.

raon0211:
What about focusing this pull request in translating static labels? We will manage other documents in a similar manner to your repository (https://github.com/anaclumos/extracranial).

anaclumos:
Sounds great. I will revert the index.md to as it is right now on the main branch.

anaclumos:
Ready for review. Please feel free to modify any JSON string translations as you wish!

One note: Docusaurus does not recognize strings outside the `docs` directory at the moment. (e.g., "라이브러리 소개"). We might want to fix this in the future!

Thank you so much, Toss team ❤️ Huge kudos to you all. Please let me know if anything needs to be done on my side!

raon0211 (approved):
LGTM 👍  Thanks for your awesome contribution, @anaclumos !

PR #33, opened 2022-10-12
fix: not use non-null assertion

ChanhyukPark-Tech:
### removed useless non-null assertion exclamation mark

in useCheckList hook listRef.current[idx] can not be null it can only be T
because listRef initialized by initialItems which should be array of T (T[])

idx is -1 only when they don't have find item in array (by findIndex function)

in line67
idx should over -1 because they found item in array they have  statement  in line 66 
so listRef.current[idx] can not be null anytime
As a result they don't have exclamation mark to confirm 

## PR Checklist

I passed all test in local


ChanhyukPark-Tech:
I realized I needed an exclamation point.
because item should be infereced  NonNullable<T> not `T extends Item` 

close my PR 

thx !

PR #36, opened 2022-10-12
test(utils): Add test code for generateID

ChanhyukPark-Tech:
## Overview

I wrote test code for generateID which belong to @toss/utils To ensure stability, I have written a test code

`nextUniqueId` which is local variable in generateID should be initialized to 0 for every test case
so I used jest.resetModules 
resetModules help our individual test code run independently from other test cases

## PR Checklist

2. I have written documents and tests, if needed. => I made test case !

### DIFF
--- a/packages/common/utils/src/generateID.spec.ts
+++ b/packages/common/utils/src/generateID.spec.ts
@@ -1,21 +1,19 @@
-describe('generateID는', () => {
-  beforeEach(() => {
-    jest.resetModules();
+import { generateID } from './generateID';
+
+describe('generateID should', () => {
+  it('return different string values each time it is called', () => {
+    const id1 = generateID();
+    const id2 = generateID();
+    const id3 = generateID();
+
+    expect(id1).not.toEqual(id2);
+    expect(id1).not.toEqual(id3);
+    expect(id2).not.toEqual(id3);
   });
 
-  it('순서대로 고유한 ID를 생성합니다.', () => {
-    import('./generateID').then(({ generateID }) => {
-      expect(generateID()).toEqual('1');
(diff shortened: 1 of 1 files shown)


raon0211:
I believe test codes should clearly show the author's intended behavior for that function.

I think what we actually intend for `generateID` is to return different `string` values each time it is called. Whether or not the ID is incremented is implementation detail. In this manner, what about changing the test code as the following code?

```typescript
it('returns different string values each time it is called', () => {
  const id1 = generateID();
  const id2 = generateID();
  const id3 = generateID();

  expect(id1).not.toEqual(id2);
  expect(id1).not.toEqual(id3);
  expect(id2).not.toEqual(id3);
});

it('should start with the prefix, if given', () => {
  const id = generateID('toss');

  expect(id.startsWith('toss')).toBe(true);
});
```

raon0211:
By writing the test code as above, the test code will become more concise and easier to read.

ChanhyukPark-Tech:
Thank you for explaining such a good example and the purpose of the test code.
I've always thought a lot about what criteria to write the test code, but you(raonraon0211) told me that the test code should be able to test the behavior that the author intended. Just by looking at the test code you wrote as an example, I understood the intention right away. 
I'll fix it and push it to this branch! Thank you.💪

raon0211 (approved):
LGTM :+1:

ChanhyukPark-Tech:
thx for solving install error!

ChanhyukPark-Tech:
The method you recommended is written in English. The other test codes were explained in Korean, so which one should I write in English or Korean?!

raon0211:
We are planning to migrate our repository into English 😓

Issue #34, opened 2022-10-12
[Feature]: common > utils > clamp error case


ChanhyukPark-Tech:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: @toss/utils

### **Overview**

In `clamp.ts`

Currently, the function is only supposed to work if bound1 is less than bound2. If user unaware of this rule and can put a larger value in bound1 than bound2. In that case, this function will not function normally. For example, if value=3, bound1=10, bound2=1, the return value must be 3, but the error results in 1.

### **Describe the solution you'd like**
i think this code works well if you agree I will make Pull request

```
export function clamp(value: number, bound1: number, bound2?: number) {
  if (bound2 == null) {
    return Math.min(value, bound1);
  }

  if (bound1 > bound2) {
    return Math.min(Math.max(value, bound2), bound1);
  }

  return Math.min(Math.max(value, bound1), bound2);
}
```

raon0211:
I guess this is in fact a bug. However, I think throwing an error early when our assumption (`bound1 <= bound2`) is broken might be a better approach. (I think no one intends to write code like `clamp(3, 10, 1)` 😄 ) What do you think?

ChanhyukPark-Tech:
Thank you for your kind answer.
This issue should have been a bug as you said, not a feature!

I think it would be better to throw an error than to give a wrong value.

I agree that no one will use it with `clamps (3, 10, 1)!`
However, the slash library is targeting a lot of users and it's getting bigger, so I think many users will use it. I think error handling is necessary because I can't be sure that it doesn't cause human errors as many users use it!

raon0211:
Yes, so I guess we should change the implementation of clamp to throw an early error.

```typescript
function clamp(value, min, max) {
  if (min > max) {
    throw new Error("min is bigger than max");
  }

  // original implementation
}

ChanhyukPark-Tech:
Yes! I agree🙋🏼‍♂️
If you allow me, I will make a PR after modifying it!
Will it be okay?

raon0211:
Yes, sure!

PR #37, opened 2022-10-12
fix(utils): error handling for clamp function

ChanhyukPark-Tech:
## Overview

If the value of the arguments in the clamp function does not inserted in the desired purpose, the error informs the user that it is an incorrect usage.

be discussed in https://github.com/toss/slash/issues/34

## PR Checklist

=> I wrote some test code for error handling!🚀

### DIFF
--- a/packages/common/utils/src/clamp.spec.ts
+++ b/packages/common/utils/src/clamp.spec.ts
@@ -16,8 +16,14 @@ describe('clamp should', () => {
   });
 
   it('throw error when bound1 is bigger than bound2', () => {
-    expect(() => clamp(3, 10, 5)).toThrowError();
-    expect(() => clamp(7, 12, 5)).toThrowError();
-    expect(() => clamp(100, 6, 5)).toThrowError();
+    expect(() => clamp(3, 10, 5)).toThrowErrorMatchingInlineSnapshot(
+      `"The value of bound2 must be a number greater than bound1."`
+    );
+    expect(() => clamp(7, 12, 5)).toThrowErrorMatchingInlineSnapshot(
+      `"The value of bound2 must be a number greater than bound1."`
+    );
+    expect(() => clamp(100, 6, 5)).toThrowErrorMatchingInlineSnapshot(
+      `"The value of bound2 must be a number greater than bound1."`
+    );
   });
 });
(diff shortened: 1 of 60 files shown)


packages/common/utils/src/clamp.spec.ts
```diff
@@ -1,17 +1,23 @@
...
     expect(clamp(10, 6, 10)).toBe(10);
     expect(clamp(6, 10, 10)).toBe(10);
     expect(clamp(7, 5, 10)).toBe(7);
     expect(clamp(100, 5, 6)).toBe(6);
   });
+
+  it('throw error when bound1 is bigger than bound2', () => {
+    expect(() => clamp(3, 10, 5)).toThrowError();
+    expect(() => clamp(7, 12, 5)).toThrowError();
+    expect(() => clamp(100, 6, 5)).toThrowError();
```
raon0211, line 21: What about using [.toThrowErrorMatchingInlineSnapshot()](https://jestjs.io/docs/expect#tothrowerrormatchinginlinesnapshotinlinesnapshot) to check whether the thrown error is our desired error?

Changing `toThrowError()` to `toThrowErrorMatchingInlineSnapshot()` and running the test will generate the inline snapshot, making the test to fail when the function throws an error that is different from the desired one

raon0211 (approved):
Thanks for your contribution 👍

packages/common/utils/src/clamp.spec.ts
```diff
@@ -1,17 +1,29 @@
 import { clamp } from './clamp';
 
-describe('clamp는', () => {
-  it('max 만 있을 때 잘 동작한다', () => {
+describe('clamp should', () => {
```
HoseungJang, line 3: it would be better to write `"should"` in `it()`:
```typescript
describe("clamp", () => {
  it("should work ...", () => {
    /* ... */
  });
});
```
so that developers can read straight.

HoseungJang: They are functionally the same but semantically different.

```typescript
it("should have a button", () => { ... }); // 1
test("should have a button", () => {}); // 2
```

you can read the code above as english sentence and then what's the more easily readable? I think `1` is the winner.

ChanhyukPark-Tech: I didn't immediately recognize the difference after seeing the comments you wrote. After reading the comments about three times, I felt exactly meaning you wrote 'semantically different', so I realized exactly that the method should be it not test. As a front-end developer, I didn't realize the difference between it and test when I wrote the test code, and I didn't get such a clear answer anywhere, but I finally got a clear answer tonight. Thanks a lot. I'll reflect on it.

PR #42, opened 2022-10-13
refactor(emotion-utils): union type reused by global type

ChanhyukPark-Tech:
## Overview

While looking at the emotion-utils project, I found a type that was used repeatedly. I thought it would be nice to declare it there and use it anywhere as there is a types.ts file in src folder . I think the bigger the project, the more effective it will be.

## PR Checklist


raon0211 (approved):
☺️  Thanks a lot for your contribution!

PR #50, opened 2022-10-14
fix(utils): rewrite test description for clamp function

ChanhyukPark-Tech:
## Overview

HoseungJang left a review on #37 Pull Request. He wanted to change the description of the test semantically and explained why. I fully understood that the reason was reasonable, and I completed the modification.
I've always thought of a way to have meaning in the code itself, I think this test description was modified well because the code itself can be intuitively understood by the developer.

## PR Checklist

=> I ran test code in local


HoseungJang (approved):
Thanks 👍

PR #52, opened 2022-10-14
fix(utils): relocate file location isDifferentArray function

ChanhyukPark-Tech:
## Overview

I think a function called `isDifferentArray` should be utils function. I moved the location of the file because I thought it would be good for many people to use it in @toss/utils, not dependent on the @toss/error-boundary package.

`question`
The name of the test file does not seem to be unified in various packages.
`*.spec.ts` `*.test.ts` which one should i take?! 
if there is any convention please tell me !

## PR Checklist


raon0211:
Yes, this change seems right. Actually we don’t have an explicit convention for test files right now. (Personally, I prefer *.spec.ts 😄) We will discuss and build a convention and let you know in this pull request. (cc. @HoseungJang @changyoungoh )

PR #53, opened 2022-10-14
docs(w3c): refactor to validate w3c  

ChanhyukPark-Tech:
## Overview

When I verified the address https://slash.page/ where this library's document is currently deployed through w3c validator, I found that there were two errors.

[url of validation](https://validator.w3.org/nu/?doc=https%3A%2F%2Fslash.page%2F)
image

so i fix two things to validate w3c.

---

1. I change autoplay="true" to just autoplay 

[mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Video)
>To disable video autoplay, autoplay="false" will not work; 
>the video will autoplay if the attribute is there in the <video> tag >at all. 

it says autoplay works just by writing autoplay itself not boolean value

---

2. I removed the style code from the inside of the article tag and put it inside the head.

If there is a reason why the developers put the style tag in the body area, I would appreciate it if you could explain it! I checked that it works well even if the style code is moved inside the head tag not in body part.

If this method is not good, I think there is a way to deploy the style code to CDN and then load it.

---

### Result

image

Since I cannot deploy these docs and test them, I modified the tag directly in the html environment and went through the verification process in w3c. I've confirmed that it passed. This is considered to have followed the correct markup rules.

## PR Checklist

### DIFF
--- a/docs/i18n/ko/docusaurus-plugin-content-pages/index.md
+++ b/docs/i18n/ko/docusaurus-plugin-content-pages/index.md
@@ -27,7 +27,7 @@ hide_table_of_contents: true
   </div>
 
   <div style={{ gridArea: 'image', textAlign: 'center' }}>
-  <video class="key-video" src="https://static.toss.im/assets/slash-libraries/keyvis.mp4" autoplay="true" muted="true" playsInline="true" loop="true" />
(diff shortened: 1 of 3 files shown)


ChanhyukPark-Tech:
its not work well in deploy preview.. i have to fix it

ChanhyukPark-Tech:
The autoplay video play and style code are not applied easily in [docusaurus](https://github.com/facebook/docusaurus) environment as I thought!
I will study more so that I can apply the design you want while passing w3c validation!

PR #59, opened 2022-10-15
docs(utils): replace any type to specified type in minBy doc

euijinkk:
## Overview

- replace any type to specified type in minBy function JSDoc

I found other mismatch between JSDoc and code. 
slash team think it should be edited? And I wonder if slash team made JSDOc one by one or not.

## PR Checklist


packages/common/utils/src/minBy.ts
```diff
@@ -7,7 +7,7 @@
  *   // 최솟값을 찾을 배열
  *   collection: T[],
  *   // 배열의 값을 계산하는 방법
- *   iteratee: (element: T) => any
+ *   iteratee: (element: T) => T | undefined
```
ChanhyukPark-Tech, line 10: I think not this should be fixed to 
```suggestion
 *   iteratee: (element: T) => number
```

according to code itself return type of iteratee function is number not T | undefined

euijinkk: Sorry, that's my mistake. 
I fixed it according to your suggestion.

[docs: fix iteratee return type](https://github.com/toss/slash/pull/59/commits/dd5395d55ee60624e61e8c9f3bed46ffbc107742)

raon0211:
Actually we manually have written all the jsdocs so there may be some mistakes 😢  thanks for your contribution!

raon0211 (approved):
Thanks a lot, @euijinkk ! 🙌

PR #60, opened 2022-10-15
fix(utils): fix mapValues type error

euijinkk:
## Overview

I found `mapValues` function has two type Error. It's because
1. T generic is not restricted. It is just unknown type. So I fixed it Record object type. But I'm not sure I should consider symbol type. First, I considered symbol, but I found slash team uses `ObjectKeys` which is for excluding symbol type. So I followed that. I wonder your opinion.
2. Object.entries typing. I fixed it by `assertion` as slash team wrote in `objectEntries`. 

## PR Checklist


packages/common/utils/src/mapValues.ts
```diff
@@ -11,8 +13,11 @@
  * mapValues({ foo: 1, bar: 2 }, x => x * 2)
  * // => { foo: 2, bar: 4 }
  */
-export function mapValues<T, U>(value: T, mapper: (value: T[keyof T]) => U): { [K in keyof T]: U } {
-  const entries = Object.entries(value);
+export function mapValues<T extends Record<PropertyKey, T[ObjectKeys<T>]>, U>(
+  value: T,
+  mapper: (value: T[keyof T]) => U
+): { [K in keyof T]: U } {
```
HoseungJang, line 19: Symbol is not enumerable so it will be excluded from the result of `Object.entries`.
So you should exclude all symbol types of the result.

packages/common/utils/src/mapValues.ts
```diff
@@ -11,8 +13,11 @@
...
+  const entries = Object.entries(value) as Array<[ObjectKeys<T>, T[ObjectKeys<T>]]>;
```
HoseungJang, line 20: This assertion refers to exactly the same type of the result of `objectEntries`.
So I think you can import and just use it.

```typescript
import { objectEntries } from './object';

const entries = objectEntries(value); // Array<[ObjectKeys<T>, T[ObjectKeys<T>]]>
```

But I think any type assertions are not required here. This function is enough to be clear and type-safe without it.

HoseungJang (changes requested):

packages/common/utils/src/mapValues.ts
euijinkk: OK. I understand. So I excluded symbol type in T generic. Is it right?

[fix: exclude symbol type in mapValue parameter](https://github.com/toss/slash/pull/60/commits/7f8abedca3d9d43bd3a1c0d4a5ee1f6452764efd)

packages/common/utils/src/mapValues.ts
euijinkk: > But I think any type assertions are not required here. This function is enough to be clear and type-safe without it.

Yes, that's right. In this case type assertion doesn't need. I removed type assertion. 

When you find  any other problems in this PR, please tell me again.

packages/common/utils/src/mapValues.ts
```diff
@@ -11,7 +13,10 @@
  * mapValues({ foo: 1, bar: 2 }, x => x * 2)
  * // => { foo: 2, bar: 4 }
  */
-export function mapValues<T, U>(value: T, mapper: (value: T[keyof T]) => U): { [K in keyof T]: U } {
+export function mapValues<T extends Record<string | number, T[ObjectKeys<T>]>, U>(
+  value: T,
+  mapper: (value: T[keyof T]) => U
+): { [K in keyof T]: U } {
```
HoseungJang, line 19: I think this isn't correct.

`extends` keyword doesn't mean 'strict equal' to passed type. You can just specify this generic should be passed as object, not the same type of `T`.

And `object.entries` accepts any object even if it has symbol keys. Symbol keys are only excluded from return value of `object.entries`.

So I think the definition of `mapValues` function should be:
```typescript
export function mapValues<T extends Record<PropertyKey, any>, U, O = Omit<T, symbol>>(
  value: T,
  mapper: (value: T[keyof O]) => U
): { [K in keyof O]: U };
```

P.S. `extends` constraint is just sub-type of the type passed into generic. So your generic accepts an object with symbol keys after all.

HoseungJang (changes requested):

packages/common/utils/src/mapValues.ts
euijinkk: I uderstand mapValues should also aceept symbol, but object entries should exclude symbol type.
Is it right? 

I include your opinion in below commit.
[fix: make mapValue get symbol type parameter](https://github.com/toss/slash/pull/60/commits/7b5f449cbea1334ff93733c49d834f4262d46aa7)

And I edit mapvValues JSDoc.
[docs: edit mapValues JSDoc](https://github.com/toss/slash/pull/60/commits/624b8c6d40e384d3a50fda0424743c17ad92a9e6)

I came into this PR to contribute your library but I learn many things about type design.
Thank you for your kind explain.

HoseungJang:
Thanks @euijinkk .

packages/common/utils/src/mapValues.ts
euijinkk: OK. 
Thank you!!!

PR #61, opened 2022-10-15
fix(utils): remove redundant template literal type in ObjectKeys

euijinkk:
## Overview

`ObjectKeys` type returns a type wrapped with template literal type. I think it is redundant. I removed it.

## PR Checklist


raon0211 (approved):
Yes this is right. Thanks @euijinkk !

PR #64, opened 2022-10-16
fix(emotion-utils): replace flex's align type with a more specific type

mass2527:
### Problem

Implementation of `flex` is slightly different from what is described in the JSDoc.

### What I did

- I changed `flex`'s alignItems type(`string`) with a more specific type(`CSSProperties['alignItems']`) for type safety and auto-completion since type changing seems to work as I expected unlike what the comment(Note) says.
-  I removed `flex`'s optional align type of JSDoc.

If there is something historical reason that it should be typed as a string, I would appreciate it if you could let me know.

## PR Checklist


HoseungJang (approved):

raon0211:
We will have look on this pull request until Tuesday, since I think it is in fact correct to skip passing `align` in the current implementation.

mass2527:
Thanks!
I look forward to hearing from you.

PR #66, opened 2022-10-17
fix(date): remove useless null check question mark in spec file

ChanhyukPark-Tech:
## Overview

in date.spec.ts

```typescript
const date = parseYYYYMMDD('2020-04-23');
    expect(date.getFullYear()).toEqual(2020);
    expect(date.getMonth()).toEqual(3);
    expect(date.getDate()).toEqual(23);
```

date cannot be null because parseYYYYMMDD return only Date type anytime unless throw error
Therefore, if the value is returned, it is absolutely a Date type so question mark is useless!

## PR Checklist

### DIFF
--- a/packages/common/date/src/date.spec.ts
+++ b/packages/common/date/src/date.spec.ts
@@ -1,12 +1,12 @@
-import { parseYYYYMMDD, getDateDistance, getDateDistanceText, TimeUnits } from './index';
+import { getDateDistance, getDateDistanceText, parseYYYYMMDD, TimeUnits } from './index';
 
 describe('parseYYYYMMDD', () => {
   test('"2020-04-23"은 2019년 4월 23일로 파싱한다.', () => {
     const date = parseYYYYMMDD('2020-04-23');
 
-    expect(date?.getFullYear()).toEqual(2020);
-    expect(date?.getMonth()).toEqual(3);
-    expect(date?.getDate()).toEqual(23);
+    expect(date.getFullYear()).toEqual(2020);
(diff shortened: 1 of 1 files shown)


ChanhyukPark-Tech:
I have question🙋🏼‍♂️ in date.spec.ts

```typescript
test('"2020-13-02"는 에러를 던진다.', () => {
    try {
      parseYYYYMMDD('2020-13-02');
    } catch (e: any) {
      expect(e.message).toBe('Invalid date format');
    }
  });

test('"2020-01-32"는 에러를 던진다.', () => {
    try {
      parseYYYYMMDD('2020-01-32');
    } catch (e: any) {
      expect(e.message).toBe('Invalid date format');
    }
 });
```

In order to test the throw error whick , you tested it using the try & catch provided by JavaScript.
What is the difference between this method and the toThrowError and toThrowErrorMatchingInlineSnapshotMatter provided by jest? I think it's better to explicitly write toThrowErrorMatchingInlineSnapshot to test the error message! Is there any special reason why you used try & catch?

Thank you for reading this long question!

raon0211:
Yes I think using `toThrow()` matchers is better because it is more explicit. Also, the original test case still succeeds when parseYYYYMMDD does not throw. What about refactoring the test cases altogether in this pull request?

ChanhyukPark-Tech:
Oh.. I was just meaningful to use matcher, but I realized that the original test code will succeed even if it is not intended. Thank you! I pushed!

raon0211 (approved):
Thanks for your awesome pull request!

PR #74, opened 2022-10-21
fix(react): added missing export in hooks/index.ts

mass2527:
## Problem
 
 Couldn't `import { useBodyClass, useResizeObserver, useStyleSheetInjection } from '@toss/react'` since it wasn't exported from hooks/index.ts
 
 ## What I did
 added missing export in hooks/index.ts
 

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/index.ts
+++ b/packages/react/react/src/hooks/index.ts
@@ -1,4 +1,5 @@
 /** @tossdocs-ignore */
+export { default as useBodyClass } from './useBodyClass';
 export { default as useBooleanState } from './useBooleanState';
 export { default as useCallbackOnce } from './useCallbackOnce';
 export { default as useCheckList } from './useCheckList';
@@ -18,7 +19,9 @@ export * from './usePreservedCallback';
 export { default as usePreservedReference } from './usePreservedReference';
 export { default as usePrevious } from './usePrevious';
 export * from './useRefEffect';
+export { default as useResizeObserver } from './useResizeObserver';
 export * from './useStorageState';
+export { default as useStyleSheetInjection } from './useStyleSheetInjection';
 export { default as useThrottle } from './useThrottle';
 export { default as useTimeout } from './useTimeout';
 export * from './useTimeoutQueue';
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks, @mass2527 !

packages/react/react/src/hooks/useResizeObserver.ts
```diff
@@ -21,7 +21,7 @@ export type OnResize = (entry: ResizeObserverEntry) => void;
  *
  * <button ref={ref}>...</button>
  */
-export function useResizeObserver<E extends HTMLElement = HTMLElement>(onResize: OnResize) {
+export default function useResizeObserver<E extends HTMLElement = HTMLElement>(onResize: OnResize) {
```
HoseungJang, line 24: This is not necessary change in this PR.

HoseungJang (changes requested):

packages/react/react/src/hooks/useResizeObserver.ts
mass2527: Thanks for your suggestion.

I already have stashed files that fix all hooks to be exported in the same way(default export).

packages/react/react/src/hooks/index.ts
```diff
@@ -14,12 +15,14 @@ export { default as useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffec
...
-export * from './useRefEffect';
-export * from './useStorageState';
+export { default as useRefEffect } from './useRefEffect';
+export { default as useResizeObserver } from './useResizeObserver';
+export { default as useStorageState } from './useStorageState';
+export { default as useStyleSheetInjection } from './useStyleSheetInjection';
 export { default as useThrottle } from './useThrottle';
 export { default as useTimeout } from './useTimeout';
-export * from './useTimeoutQueue';
+export { default as useTimeoutQueue } from './useTimeoutQueue';
```
HoseungJang, line 27: I think default export looks inefficient because we should name it twice.
How about named export?

```typescript
// useSomething.ts
export default function useSomething() { ... }

// index.ts
export { default as useSomething } from "./useSomething";
```

mass2527: I totally agree with you. I'll change all hooks with named export!
I thought default export was a convention since most hooks are exported that way.

ChanhyukPark-Tech:
This discussion was so interesting that I also changed the components folder! Thank you. [mass2527](https://github.com/mass2527) and [HoseungJang](https://github.com/HoseungJang)

PR #77, opened 2022-10-22
fix(react): replace all default export of components with named export

ChanhyukPark-Tech:
## Overview

I saw an interesting discussion at #74 ! I have always used default export, but it was inconvenient that I had to write function name twice, but I realized it because HoseungJang reviewd exactly.
so I replace all default export component with named export

## PR Checklist


changyoungoh (approved):
LGTM 👍

PR #78, opened 2022-10-22
test(react): write OpenGraph test code

ChanhyukPark-Tech:
## Overview

OpenGraph, which sets the meta tag, which is such a good component, does not have a test code, so I wrote it for safety .
when merge #77 it should fix something just import part

## PR Checklist

2. I have written documents and tests, if needed. yes I made them


packages/react/react/src/components/OpenGraph/index.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+    expect(getMeta('og:description')).toEqual('금융의 모든 것, 토스에서 쉽고 간편하게');
+  });
+  it(`should render inserted imageUrl`, () => {
+    render(<OpenGraph imageUrl="https://static.toss.im/homepage-static/newtoss/newtoss-og.jpg" />);
+    expect(getMeta('og:image')).toEqual('https://static.toss.im/homepage-static/newtoss/newtoss-og.jpg');
+  });
+  it(`should throw error when no meta tag is found `, () => {
+    render(<OpenGraph />);
+    expect(() => getMeta('og:title')).toThrowError();
+  });
```
ChanhyukPark-Tech, line 29: I implemented it like this to test the error part, is it the right method? If you have a better way, please recommend it!

packages/react/react/src/components/OpenGraph/index.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+    for (let i = 0; i < metas.length; i += 1) {
+      if (metas[i].getAttribute('property') === metaName) {
+        return metas[i].getAttribute('content');
+      }
+    }
+    throw new Error('cannot find your meta tag');
+  };
+  it(`should render inserted title`, () => {
+    render(<OpenGraph title="토스" />);
+    expect(getMeta('og:title')).toEqual('토스');
```
manudeli, line 16:
```suggestion
    const title = '토스'

    render(<OpenGraph title={title} />);
    expect(getMeta('og:title')).toEqual(title);
```

packages/react/react/src/components/OpenGraph/index.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+  });
+  it(`should render inserted description`, () => {
+    render(<OpenGraph description="금융의 모든 것, 토스에서 쉽고 간편하게" />);
+    expect(getMeta('og:description')).toEqual('금융의 모든 것, 토스에서 쉽고 간편하게');
```
manudeli, line 20:
```suggestion
    const description = '금융의 모든 것, 토스에서 쉽고 간편하게'

    render(<OpenGraph description={description} />);
    expect(getMeta('og:description')).toEqual(description);
```

packages/react/react/src/components/OpenGraph/index.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+  });
+  it(`should render inserted imageUrl`, () => {
+    render(<OpenGraph imageUrl="https://static.toss.im/homepage-static/newtoss/newtoss-og.jpg" />);
+    expect(getMeta('og:image')).toEqual('https://static.toss.im/homepage-static/newtoss/newtoss-og.jpg');
```
manudeli, line 24:
```suggestion
    const imageUrl = 'https://static.toss.im/homepage-static/newtoss/newtoss-og.jpg'
    
    render(<OpenGraph imageUrl={imageUrl} />);
    expect(getMeta('og:image')).toEqual(imageUrl);
```

manudeli (commented):
It's really trivial, but How about below suggestion comments to avoid chance that 2 long strings can be slightly different by human mistake not on purpose?

your intention is that OpenGraph's title is ```equal``` with Meta's title
So I want anyone cannot make mistake if new test code is required

packages/react/react/src/components/OpenGraph/index.spec.tsx
ChanhyukPark-Tech: Your suggestion is so clear
it cannot make human mistake
I will apply your suggestion 
Thank you!

packages/react/react/src/components/OpenGraph/index.spec.tsx
ChanhyukPark-Tech: thx!
I applied it!

packages/react/react/src/components/OpenGraph/index.spec.tsx
raon0211: I would probably return `null` when errors can happen quite often, but this is really a minor opinion.

raon0211 (approved):
Thanks, @ChanhyukPark-Tech !

PR #80, opened 2022-10-23
fix(framer-motion): change directory

manudeli:
## How about change directory for framer-motion into react from common?
Current position of @toss/framer-motion, common can make us hard to find it
When I found directory of @toss/framer-motion, I was confused because of directory

Below picture is description of original framer-motion main homepage.
> A production-ready ```motion library for React```. Utilize the power behind Framer, the best prototyping tool for teams. Proudly open source.

framer-motion is library for react
so, How about changing directory of @toss/framer-motion into react from common

스크린샷 2022-10-23 15 21 28

## PR Checklist


changyoungoh (approved):
LGTM 👍

PR #82, opened 2022-10-23
fix(async-boundary): add generic constraint of withAsyncBoundary

manudeli:
# AS-IS
image

## before this PR, Props can be any type like string.
but I thought string or number or Function or Array is not proper type for Props

image

# TO-BE
Constraint by Record<string, unknown>

```typescript
export default function withAsyncBoundary<Props extends Record<string, unknown> = Record<string, never>>(
  WrappedComponent: ComponentType<Props>,
  asyncBoundaryProps: ComponentProps<typeof AsyncBoundary>
) {
  return (props: Props) => {
    return (
      <AsyncBoundary {...asyncBoundaryProps}>
        <WrappedComponent {...props} />
      </AsyncBoundary>
    );
  };
}
```

image

## PR Checklist


packages/react/async-boundary/src/withAsyncBoundary.tsx
```diff
@@ -17,7 +17,7 @@ import AsyncBoundary from './AsyncBoundary';
  * @see https://jbee.io/react/error-declarative-handling-1/ 선언적으로 로딩과 에러 상태 처리하기
  * @see https://toss.im/slash-21/sessions/3-1 Suspense와 에러 처리 관련된 Slash 21 발표
  */
-export default function withAsyncBoundary<Props = {}>(
+export default function withAsyncBoundary<Props extends Record<string, unknown> = Record<string, never>>(
```
HoseungJang, line 20: I think any constraints are not required here because all React developers might know props should be object, and even if other types were passed it wouldn't work. I think that's why React doesn't provide any constraint in `ComponentType` too.

But, apart from that, this change looks good to me.

HoseungJang (approved):

HoseungJang:
Thanks @manudeli .

PR #83, opened 2022-10-23
fix(emotion-utils):  refactor position function

yoonminsang:
## Overview

- improve type safe
- change wrong variable name

## PR Checklist

- [x] I read and included theses actions below

1. I have read the [Contributing Guide](https://github.com/toss/slash/blob/main/.github/CONTRIBUTING.md)
2. I have written documents and tests, if needed.

PR #84, opened 2022-10-24
fix(emotion-utils): refactor position function 

yoonminsang:
## Overview

- improve type safe
- change wrong variable name
- change nested ternaries to IIFE

https://github.com/toss/slash/pull/83

## PR Checklist


packages/react/emotion-utils/src/utils/position.ts
```diff
@@ -71,14 +71,23 @@ export function position(position: CSSProperties['position'], coordinates?: Coor
...
+  topOrRightOrCoordinates: CSSPixelValue | Coordinates = {},
   ...values: CSSPixelValue[]
 ) {
-  const [top, right, bottom, left] = isPositionValue(positionOrTop)
-    ? isCSSPixelValue(topOrCoordinates)
-      ? [topOrCoordinates, ...values]
-      : [topOrCoordinates?.top, topOrCoordinates?.right, topOrCoordinates?.bottom, topOrCoordinates?.left]
-    : [positionOrTop, topOrCoordinates as CSSPixelValue, ...values];
+  const [top, right, bottom, left] = (() => {
+    if (!isPositionValue(positionOrTop)) {
```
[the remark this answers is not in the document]
robin-classum: okey. I add some comments. Thank you~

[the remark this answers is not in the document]
yoonminsang: okey. I add some comments. Thank you~

packages/react/emotion-utils/src/utils/position.ts
```diff
@@ -71,14 +71,26 @@ export function position(position: CSSProperties['position'], coordinates?: Coor
...
+    // position(position, top, right, bottom, left);
+    if (!isCSSPixelValue(topOrRightOrCoordinates)) {
+      return [
+        topOrRightOrCoordinates.top,
+        topOrRightOrCoordinates.right,
+        topOrRightOrCoordinates.bottom,
+        topOrRightOrCoordinates.left,
+      ];
+    }
+    // position(position, coordinates);
```
HoseungJang, line 91: It's reversed.

yoonminsang: ah... Sorry to bother you.

packages/react/emotion-utils/src/utils/position.ts
```diff
@@ -87,7 +99,8 @@ export function position(
 }
 
 function isPositionValue(value: unknown): value is Property.Position {
-  return [`static`, `relative`, `absolute`, `fixed`, `sticky`, `-webkit-sticky`].includes(value as string);
+  const positions: Property.Position[] = ['static', 'relative', 'absolute', 'fixed', 'sticky', '-webkit-sticky'];
```
HoseungJang, line 102: What is the defference?

packages/react/emotion-utils/src/utils/position.ts
```diff
@@ -87,7 +99,8 @@ export function position(
...
+  return positions.includes(value as Property.Position);
```
HoseungJang, line 103: `value` is NOT `Property.Position`.

HoseungJang (changes requested):

packages/react/emotion-utils/src/utils/position.ts
yoonminsang: That's right. I don't want to make a mistake this time, so I read code again.

I don't know why position function use `CSSProperties['position']`. Is there any other reason?
ex) next code is type safe.
```position(undefined, 1, 1, 1, 1)```

How about this code? 

```
type PositionOrTop = Property.Position | CSSPixelValue;
export function position(
  position: Property.Position,
  top: CSSPixelValue,
  right: CSSPixelValue,
  bottom: CSSPixelValue,
  left: CSSPixelValue,
): SerializedStyles;
export function position(
  top: CSSPixelValue,
  right: CSSPixelValue,
  bottom: CSSPixelValue,
  left: CSSPixelValue,
): SerializedStyles;
export function position(position: Property.Position, coordinates?: Coordinates): SerializedStyles;

export function position(
  positionOrTop: PositionOrTop,
  topOrRightOrCoordinates: CSSPixelValue | Coordinates = {},
  ...values: CSSPixelValue[]
) {
 // ...
}

function isPositionValue(value: PositionOrTop): value is Property.Position {
  return typeof value === 'string';
}
```

Comment (HoseungJang) [reply packages/react/emotion-utils/src/utils/position.ts:103]: That makes sense. It's a bug that the first parameter of `position` can be `undefined`.
So please fix the type of it to `Property.Position`.

But please keep the implementation of `isPositionValue` because `CSSPixelValue` also can be string.

Comment (yoonminsang) [reply packages/react/emotion-utils/src/utils/position.ts:103]: I miss again...

This code is not working in typescript.

```
function isPositionValue(value: PositionOrTop): value is Property.Position {
  const positions: Property.Position[] = ['static', 'relative', 'absolute', 'fixed', 'sticky', '-webkit-sticky'];
  return positions.includes(value);
}
```
스크린샷 2022-10-25 오후 11 57 04

스크린샷 2022-10-25 오후 11 35 42

So I wrote the code like this. I think it doesn't matter because value has type. But I don't know if you agree. What do you think about it?

Other ways
- use ts-ignore
- `positions.some((position) => position === value);`

```
  const positions: Property.Position[] = ['static', 'relative', 'absolute', 'fixed', 'sticky', '-webkit-sticky'];
  return positions.includes(value as Property.Position);
```

Comment (HoseungJang) [reply packages/react/emotion-utils/src/utils/position.ts:103]: ?? I have already told you `value` is NOT `Property.Position`. Just keep the previous implementation of `isPositoinValue`.

It's the role of `isPositionValue` to determine that `value` is `Property.Position`.

```typescript
function isPositionValue(value: unknown): value is Property.Position {
  return ['static', 'relative', 'absolute', 'fixed', 'sticky', '-webkit-sticky'].includes(value as any);
}
```

yoonminsang: okey.. i understand

yoonminsang: Thanks for commenting. It was time to learn about typescript and toss library.

HoseungJang (approved):

PR #85, opened 2022-10-24
fix: remove useDidUpdate hook

mass2527:
### Conclusion
 **`useDidUpdate` hook should be removed since, in the long run, it’s likely to introduce more problems than it solves.**
 
 ### Reason
 
 According to the new react docs, react team says

"**Keep custom Hooks focused on concrete high-level use cases.** Avoid creating and using custom “lifecycle” Hooks that act as alternatives and convenience wrappers for the useEffect API itself"

They think that custom "lifecycle" hooks like `useMount`, `useEffectOnce`, `useUpdateEffect` don't fit well into the React paradigm. Because linter won't warn about the wrong dependency(linter has no idea about this kind of hook) and this will lead to lots of bugs in the long run.

### Reference

[When to use custom hooks](https://beta.reactjs.org/learn/reusing-logic-with-custom-hooks#when-to-use-custom-hooks) DEEP DIVE section - Keep your custom Hooks focused on concrete high-level use cases (still beta, but worth considering it)

## PR Checklist


ChanhyukPark-Tech:
Thank you for sharing good knowledge.
If you work on the project with React, don't you want useEffect to not work when you first mount it? So I used to use this useDidUpdate hook. In that case, is it the best option to avoid that situation?

HoseungJang:
This is completely breaking change. You **SHOULDN'T REMOVE** any modules carelessly for someone using it.

And I don't think the new React documentation says "Don't use any 'lifecycle' custom hooks". I think they say "Be careful about it".

Assuming that you don't want to trigger `useEffect` callback when dependencies are changed, and then you write the same code as `useDidUpdate`.

As you said, there's no linter warning even if `useEffect` deps are missing, it's very dangerous. But if it's intended behavior, it's not dangerous.

mass2527:
Thanks for sharing the toss team's opinion on this!

I just wanted to tell maintainers, from my point of view, it would be better to make it whenever needed.

```ts
  const hasMounted = useRef(false);

  useEffect(() => {
    if (!hasMounted.current) {
      hasMounted.current = true;
      return;
    }

    // write effect here (run whenever dependencies change)
  }, dependencies); // linter blames the wrong dependencies
```

Although both ways have tradeoffs,
- using custom life cycle hooks is convenient but no linter warnings
- not using custom life cycle hooks is inconvenient but the linter tells you what's wrong

I guess the latter is slightly better than the former, because, this is a public library for everyone.
So we need to suppose that there might be someone who doesn't have enough knowledge of the effect's dependencies.

PR #88, opened 2022-10-25
chore: Add nvm configuration

manudeli:
## .nvmrc for nvm user using [deeper shell integration](https://github.com/nvm-sh/nvm#deeper-shell-integration)
Many JavaScript developers may use [nvm](https://github.com/nvm-sh/nvm) to change node.js version rapidly.

Not perfectly, but It'll partially prevent contributors can use too old node.js version
In my case, nvm default node version is 14. but If this project have .nvmrc in root, my zsh-hook will change my node.js version to 16.17 automatically by using [deeper shell integration](https://github.com/nvm-sh/nvm#deeper-shell-integration).

[Shopify/quilt](https://github.com/Shopify/quilt/blob/main/.nvmrc), [tanstack/query](https://github.com/TanStack/query/blob/main/.nvmrc) are using .nvmrc too. I think it's partially good way to prevent chance to make mistake. and I want to suggest to @toss/slash

@toss/slash might [scope at node.js version 16.17](https://github.com/toss/slash/blob/main/.circleci/config.yml#L59) currently. So I added 16.17 into .nvmrc

## PR Checklist


.nvmrc
```diff
@@ -0,0 +1 @@
+16.17
```
HoseungJang, line 1: (Minor)Add new line please.

HoseungJang (changes requested):

.nvmrc
raon0211, line 1:
```suggestion
16.17

```

raon0211 (approved):
Thanks, @manudeli !

PR #91, opened 2022-10-27
fix(storage): optimize function call in useIsomorphicLayoutEffect hook

ChanhyukPark-Tech:
## Overview

The `isServer` is being called within the `isClient` function, which thought an unnecessary function call was occurring once.
So I made the isServer function call directly from inside the useIsomorphicLayoutEffect hook.

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useIsomorphicLayoutEffect.ts
+++ b/packages/react/react/src/hooks/useIsomorphicLayoutEffect.ts
@@ -1,15 +1,15 @@
-import { isClient } from '@toss/utils';
+import { isServer } from '@toss/utils';
 import { useEffect, useLayoutEffect } from 'react';
 
 /**
  * @description
- * Client-side에서는 useLayoutEffect 방식을 쓰고, Server-side에서는 useEffect 방식을 쓰도록 하는 hook 입니다.
- *
  * Server-side에서는 useLayoutEffect 함수를 사용하면 warning 오류가 발생하기 때문에 사용하는 함수입니다.
+ *
+ * Client-side에서는 useLayoutEffect 방식을 쓰고, Server-side에서는 useEffect 방식을 쓰도록 하는 hook 입니다.
  *
  * @example
  * useIsomorphicLayoutEffect(() => {
  *   setSwipeRefreshEnabled(false);
  * }, []);
  */
-export const useIsomorphicLayoutEffect = isClient() ? useLayoutEffect : useEffect;
+export const useIsomorphicLayoutEffect = isServer() ? useEffect : useLayoutEffect;


raon0211:
I agree in that this may result in fewer function calls, but I don't think this is meaningful, since modern compilers can easily optimize the results. ([terser repl](https://try.terser.org/)) I wish we could focus on more meaningful optimizations,  test additions, or bug fixes.

image

What's your opinion on this?

ChanhyukPark-Tech:
I got it! I'm fully understand what you said
It was my first time to know that there was a good tool called Terser and I was able to have time to study!

I should test the source code that I'm curious about the results of the build on terser Thank you

ChanhyukPark-Tech:
@mooyoul 

Thank you for kindly writing a review on the closed pull request!
I don't think this hook will be called much in the react life cycle! As you said, the hardware has improved these days, so I think the slight optimization will be hard to affect!

The purpose of this pull request is not only to optimize, but also to improve the readability and developer experience (I think 😅).

1. Developers should check the `isClient` function to see what `useIsomorphicLayoutEffect` is and enter the `isServer` to check the real logic to see how it is implemented. That's what I did. I thought that only as isServer would prevent the developer from entering depth 2! in their editor or github repository

2. I understand that SSR techniques are used a lot to provide a good user experience in modern society. I think this code is often used on the server. That's why I thought branching into semantically isServer could meaningfully give developer a better developer experience!

Once again, thank you for the great review. Future when I think of optimization, is it really necessary to think of this review? I'll have a habit of thinking about it.

Have a great day!💪

HoseungJang:
@ChanhyukPark-Tech 

- Assigning appropriate effect hook to `useIsomorphicLayoutEffect` will be occurred once at the global scope.

- > Developers should check the isClient function to see what useIsomorphicLayoutEffect is and enter the isServer to check the real logic to see how it is implemented.
    - As you say, **developers should check `isServer` to see what `useIsomorphicLayoutEffect` is too.**
    - `useIsomorphicLayoutEffect` is a hook to use `useLayoutEffect` isomorphically on both server-side and client-side, because using `useLayoutEffect` on server-side occurs a warning.
        - Then, what's the relationship between checking the implementation of `isClinet` and `isServer` out and what `useIsomorphicLayoutEffect` is? **THEY ARE NOT RELATED.**
- > I understand that SSR techniques are used a lot to provide a good user experience in modern society. I think this code is often used on the server. That's why I thought branching into semantically isServer could meaningfully give developer a better developer experience!
    - Think differently. Semantically `isClient` gives the same developer experience because `SSR techniques are used a lot to provide a good user experience in modern society`.
    - This code is used on not only the server but also the client. And See the example below. Is there any improvement in developer experience between `before` and `after`? You did the same job.
    ```typescript
    // before
    a > b ? a : b;

    // after
    b <= a ? b : a;
    ```

ChanhyukPark-Tech:
@HoseungJang 
Thank you for the good explanation!
I understand everything you said!😁

> As you say, developers should check isServer to see what useIsomorphicLayoutEffect is too.

What I was talking about was the depth to see the actual implementation of the code.
Suppose you check the code on GitHub.
To determine what useIsomorphicLayoutEffect is, 

enter: https://github.com/toss/slash/blob/main/packages/react/react/src/hooks/useIsomorphicLayoutEffect.ts

A function called isClient is used, so to check what is isClient, go to the following address.

https://github.com/toss/slash/blob/main/packages/common/utils/src/device/isClient.ts

I've been entered at this address However, the actual implementation cannot be determined yet. A function called isServer is used, so go to the following address to check.

https://github.com/toss/slash/blob/main/packages/common/utils/src/device/isServer.ts

It's only now that developers can see the real implementation.

The code I wrote was that I told you that I could improve the developer experience in terms of saving users the trouble of going to the https://github.com/toss/slash/blob/main/packages/common/utils/src/device/isClient.ts address to check what isClient is😁

HoseungJang:
@ChanhyukPark-Tech  Your point is incorrect. I mean you don't have to see the implementation of `isClient` to see what `useIsomorphicLayoutEffect` is in the first place. But why do you still consider the depth? Why do you wonder what `isClient` is in this case?

You can understand that `useIsomorphicLayoutEffect` is `useEffect` on server-side, and `useLayoutEffect on client-side without checking the implementation of `isClient`.

Considering the case to see the code in Github IS NOT MEANINGFUL improvement of developer experience. In the first place seeing the code in the Github basically has a BAD developer experience.

ChanhyukPark-Tech:
Thank you! 
I knew from you
1. declarative thinking (don't have to see the implementation)
2. seeing the code on Github is a bad situation

I've learned so much from this conversation, not just code, but how to think Thank you😎

PR #95, opened 2022-10-27
feat(react): add options in useDebounce hook

ChanhyukPark-Tech:
## Overview

useDebounce hook use the debounce function provided by the lodash, I tried to follow the implementation as much as possible and implemented it so that we could give an options to useDebounce hook. that transfer to debounce

I refer [useThrottle hook](https://slash.page/libraries/react/react/src/hooks/useThrottle.ts.tossdocs)

## PR Checklist


packages/react/react/src/hooks/useDebounce.ts
```diff
@@ -1,4 +1,4 @@
-import debounce from 'lodash/debounce';
+import { debounce, type DebounceSettings } from 'lodash';
```
HoseungJang, line 1: Please use only `lodash/debounce` for bundling optimization.
See [per-method-packages](https://lodash.com/per-method-packages).

HoseungJang (changes requested):

packages/react/react/src/hooks/useDebounce.ts
HoseungJang: You can use `Parameters` to extract the type of options.
https://github.com/toss/slash/blob/9b7ee8a03fc719a56858b7c16e4ff37a11b1cc0e/packages/react/react/src/hooks/useThrottle.ts#L18

ChanhyukPark-Tech: thank you for comment about bundling optimization!
and I used Utility type! 💪

ChanhyukPark-Tech: [fix: @toss/react useDebounce build optimization](https://github.com/toss/slash/pull/95/commits/13e91dd2da9b0cc91ee48ff92659ec0a2393081b)
[13e91dd](https://github.com/toss/slash/pull/95/commits/13e91dd2da9b0cc91ee48ff92659ec0a2393081b)

packages/react/react/src/hooks/useDebounce.ts
```diff
@@ -12,12 +12,16 @@ import { usePreservedCallback } from './usePreservedCallback';
...
+  wait: number,
+  options?: Parameters<typeof debounce>[2]
+) {
   const preservedCallback = usePreservedCallback(callback);
 
   const debounced = useMemo(() => {
-    return debounce(preservedCallback, wait);
-  }, [preservedCallback, wait]);
+    return debounce(preservedCallback, wait, options);
+  }, [options, preservedCallback, wait]);
```
HoseungJang, line 24: A reference of `options` parameter can be changed every render because developers may use `useDebounce` like:
```typescript
// passes options object directly
const debounced = useDebounce(callback, { ... });
```

But A debounced callback generated by `lodash/debounce` should keep its own reference. Otherwise debouncing doesn't work as you intended.

So please apply `usePreservedReference` to `options` parameter.
```typescript
import { usePreservedReference } from "./usePreservedReference";

const preservedOptions = usePreservedReference(options);
...
```

HoseungJang (changes requested):

packages/react/react/src/hooks/useDebounce.ts
```diff
@@ -12,12 +12,16 @@ import { usePreservedCallback } from './usePreservedCallback';
  *   getV2Logger().log(schemaId, parameter);
  * }, 500);
  */
-export function useDebounce<F extends (...args: any[]) => any>(callback: F, wait: number) {
+export function useDebounce<F extends (...args: any[]) => any>(
+  callback: F,
+  wait: number,
+  options?: Parameters<typeof debounce>[2]
```
HoseungJang, line 18: My recommendation is using [default function parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).
```typescript
function useDebounce<F extends (...args: any[]) => any>(
  callback: F,
  wait: number,
  options: Parameters<typeof debounce>[2] = {}
);
```

packages/react/react/src/hooks/useDebounce.ts
HoseungJang: https://github.com/toss/slash/pull/96

ChanhyukPark-Tech: Thank you so much for explaining it with an example!
I will take a closer look at the codes for the hooks used and apply them!
Thank you again.🚀

ChanhyukPark-Tech: You explained it in detail, but I don't think I understood it well.
If you don't mind, I'd like to ask you a question.

```typescript
const [trailing, setTrailing] = useState<boolean>(false);

const debounced = useDebounce(callback, { 
   leading: false;
   trailing: trailing;
});

{...}

setTrailing(true);
```

Are you saying that the second argument of useDebounce is the same reference even though the trailing has changed from false to true value in the above code situation, so debounced callback trailing options is still false? 

If the above situation is correct, I will completely agree with the suggestion and modify the code!

HoseungJang: No. I have said:
> A reference of options parameter can be changed on every render

HoseungJang: See the code below.
```typescript
// A
const options = useMemo(() => ({ leading: false, trailing: trailing }), [trailing]);
// B
const options = { leading: false, trailing: trailing };
```
Do you know the difference between `A` and `B`?

HoseungJang: - `A`
    - `options` object will be newly created only if `trailing` is changed. Even if the component re-rendered, `options` wouldn't be changed without change of `trailing`.
- `B`
    - `options` object will be newly created on every re-render regardless change of its properties.

ChanhyukPark-Tech: yes I know what useMemo situate

If I don't wrap it with useMemo, the reference of the object can change for each rendering due to a change in the other state, so if the options I sent you didn't actually change, you don't have to receive a new debounced callback

I got it perfectly! I'll fix it!

packages/react/react/src/hooks/useDebounce.ts
ChanhyukPark-Tech: thx! I learned from you
[fix: @toss/react useDebounce options use preservedOptions](https://github.com/toss/slash/pull/95/commits/cf56a618e3c87a3e34c166b30d26afe41f4b9548)
[cf56a61](https://github.com/toss/slash/pull/95/commits/cf56a618e3c87a3e34c166b30d26afe41f4b9548)

packages/react/react/src/hooks/useDebounce.ts
HoseungJang: Then you can ask me, "As you say, Developers memoize parameters, don't they?"

But, for developers, it's a good interface to make them pass parameters without considering memoization.

```typescript
useDebonce(setInput, 500, { leading: false }); // good
useDebounce(setInput, 500, useMemo(() => ({ leading: false }), [])); // bad. it's too verbose.
```

So I have suggested you to keep the current interface and apply `usePreservedReference`.

ChanhyukPark-Tech: As expected, you value the experience of a developer I learned a lot!
and thanks for good examples

HoseungJang (approved):
Thanks @ChanhyukPark-Tech .

Issue #103, opened 2022-10-29
[BUG]: commaize not working in RN Android


201411108:
### **Package Scope** 

Package name: @toss/utils@1.0.4

### **Describe the bug**

`commaizeNumber` didn't work at Android which written by RN.

The function returns `String(value).replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1,')` when input argument value's type is `string`, but, it's weired that convert string type variants to string.

### **Expected behavior**

Here is my code.

```typescript
const SOME_NUMBER = 1000000;

return (
    <Text>{commaizeNumber(SOME_NUMBER}/<Text>
);
```

I expected it works as same in Android and iOS, but it didn't work at Android.

* iOS case
![iOS](https://user-images.githubusercontent.com/37533787/198837703-6db627b8-51a0-4f74-9a9e-cbc61eaddce3.jpg)

* Android case
<img width="379" alt="Android" src="https://user-images.githubusercontent.com/37533787/198837727-9d630868-33b2-4216-b1f2-78f44b9a77c4.png">

It works when I change code above like below.

```typescript
const SOME_NUMBER = 1000000;

return (
    <Text>{commaizeNumber(String(SOME_NUMBER)}/<Text>
);
```

### **To Reproduce** 

Change `commaizeNumber` functions as below. Delete unnecessary type conversion. Converting value only when value type is number.

```typescript
export function commaizeNumber(value: string | number) {
  if (typeof value === 'number') {
    return String(value).replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1,');
  }

  return value.replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1,');
}
```

### **Possible Solution** 

I made PR to solve this issue #104 .

201411108:
closed with #104

PR #104, opened 2022-10-29
fix(utils): fix commaize not working in RN

201411108:
## Overview

I solved issue #103 and refactor converting same type issues.

## PR Checklist


packages/common/utils/src/Numbers.ts
```diff
@@ -74,13 +74,7 @@ export function formatToKRW(
...
 
 export function commaizeNumber(value: string | number) {
-  if (typeof value === 'string') {
-    return String(value).replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1,');
-  }
-
-  // maximumFractionDigits은 최대 소수점 자리수인데, default가 3입니다.
-  // 3자리보다 많은 수를 쓰는 경우가 있을 수도 있겠다고 생각해서 넉넉하게 10으로 잡았습니다.
-  return value.toLocaleString('en-us', { maximumFractionDigits: 10 });
+  return String(value).replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1,');
```
HoseungJang, line 77: This will make the result below:

```typescript
commaizeNumber('0.1234'); // '0.1,234'
```

How about split `value` by `.`?

HoseungJang (changes requested):

packages/common/utils/src/Numbers.ts
HoseungJang:
```typescript
const numStr = String(value);
const decimalPointIndex = numStr.indexOf('.');

if (decimalPointIndex > -1) {
  ...
} else {
  ...
}
```

201411108: I fixed this code and check it works!
I checked other functions which using `commaizeNumber`. But, there are any edge cases can existed, so, I asked you for cross-check. Thanks for recommendation!

HoseungJang: How about adding some test cases to prove that `commaize` works correctly?

HoseungJang:
```typescript
test('commaizeNumber', () => {
  expect(commaizeNumber(1234.1234)).toEqual('1,234.1234');
  expect(commaizeNumber(100)).toEqual('100');
  expect(commaizeNumber(13209802)).toEqual('13,209,802');
  expect(commaizeNumber('1234.1234')).toEqual('1,234.1234');
  expect(commaizeNumber('100')).toEqual('100');
  expect(commaizeNumber('13209802')).toEqual('13,209,802');
});
```

201411108: I added some test cases and it works well!
Thank you!

HoseungJang (approved):
Thanks @201411108 !

PR #105, opened 2022-10-30
fix: add `@types/jest` to the root

HoseungJang:
## Overview

There are type errors that cannot resolve `jest` utils in all test files.

`tsconfig.json` in the root doesn't exclude any test files. But, in the package, it's excluded by extended `tsconfig.build.json`.

So I added `@types/jest` to root to avoid the errors.

## PR Checklist


raon0211 (approved):
Thanks!

PR #106, opened 2022-10-30
fix(react): specify type for dependency array

ChanhyukPark-Tech:
## Overview

In useRefEffect deps type is just `any[]` 

image

It doesn't fit the type of useCallback.
strengthened the type and made it stable.

## PR Checklist

2. I have written documents and tests, if needed. yes passed


raon0211 (approved):
Thanks!

PR #111, opened 2022-11-02
fix(utils): fix incorrect example in docs & add test code

ChanhyukPark-Tech:
## Overview

- Docs example for decommaizeNumber was given as an example for commaizeNumber. 
  - [fix: @toss/utils decommaizeNumber docs](https://github.com/toss/slash/commit/1e5d1f716d5e1943dcf6f900bc80a2d9775cc1bc)
- There are no test codes created for decommaizeNumber, so I wrote some.
  -   [feat: @toss/utils write decommaizeNumber test code](https://github.com/toss/slash/commit/e7c04c1abb1b8599d96dd99349e123112394fb12)

## PR Checklist

### DIFF
--- a/packages/common/utils/src/Numbers.spec.ts
+++ b/packages/common/utils/src/Numbers.spec.ts
@@ -40,7 +40,7 @@ describe('Numbers', () => {
 
   test('decommaizeNumber', () => {
     expect(decommaizeNumber('13,209,802')).toEqual(13209802);
-    expect(decommaizeNumber('1234,1,,234')).toEqual(12341234);
+    expect(decommaizeNumber('12,341,234')).toEqual(12341234);
   });
 
   test('formatToKoreanNumber', () => {
--- a/packages/common/utils/src/Numbers.spec.ts
+++ b/packages/common/utils/src/Numbers.spec.ts
@@ -1,6 +1,7 @@
 import {
   ceilToUnit,
   commaizeNumber,
+  decommaizeNumber,
   floorToUnit,
   formatBusinessRegistrationNumber,
   formatPhoneNumber,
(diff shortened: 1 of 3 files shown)


packages/common/utils/src/Numbers.spec.ts
```diff
@@ -37,6 +38,11 @@ describe('Numbers', () => {
     expect(commaizeNumber('100000000')).toEqual('100,000,000');
   });
 
+  test('decommaizeNumber', () => {
+    expect(decommaizeNumber('13,209,802')).toEqual(13209802);
+    expect(decommaizeNumber('1234,1,,234')).toEqual(12341234);
```
HoseungJang, line 43: Fix this please.

```typescript
expect(decommaizeNumber('12,341,234')).toEqual(12341234);
```

HoseungJang (changes requested):

packages/common/utils/src/Numbers.spec.ts
ChanhyukPark-Tech: I wanted to show that it works well even if there are multiple comma in a string, so is there no need to verify it?🙋🏼‍♂️

raon0211: I think this is not our intended behavior (we expect only string literals which are correctly commaized), so I think excluding this from our test case is better. (Test cases should only include what we intend the function to do.)

ChanhyukPark-Tech: Thank you for your answer!
Only the test code was wrote to verify the intention of the function author.
It was an opportunity to think about it again Thank you so much!

HoseungJang (approved):
Thanks @ChanhyukPark-Tech  @raon0211 . Next time I will comment with more details.

PR #113, opened 2022-11-03
docs: Translate some docs

youngjae99:
## Overview

Migrating and translating some jsdoc to i18n documentations

## PR Checklist


packages/common/hangul/src/utils.en.md
```diff
@@ -0,0 +1,40 @@
+---
+title: utils
+---
+
+# hasBatchim
+
+Check if a hangul string has a [final consonants (batchim)](https://en.wikipedia.org/wiki/Hangul_consonant_and_vowel_tables) at the end.
```
HoseungJang, line 7: `hangul/src/utils.ts` is not exported from `@toss/hangul` so I think it doesn't have to be added to our documentation.

HoseungJang (changes requested):

packages/common/utility-types/src/serializable.ts
```diff
@@ -1,11 +1,5 @@
 type Primitive = undefined | null | boolean | number | symbol | string;
```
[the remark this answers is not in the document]
youngjae99: > If you don't use jsdoc, I think you'd better add this comment! at top of this file📌 `/** @tossdocs-ignore */`

My mistake. Thanks for checking!

packages/common/utils/src/reverseKeyValue.en.md
```diff
@@ -0,0 +1,16 @@
+# reverseKeyValue
+
+Returns an object in the form of { [value]: key } by replacing key and value
```
raon0211, line 3:
```suggestion
Returns an object in the form of { [value]: key } by swapping key and value
```

raon0211 (approved):
Thanks, @youngjae99 !

PR #116, opened 2022-11-03
docs(utils): in range function end value can be optional

ChanhyukPark-Tech:
### In docs

image

If you check the implementation of the range function below, the end parameter is optional. So I modified docs.

```tsx
// range.ts
export function range(start: number, end?: number, step = 1): number[] {
  const _start = end === undefined ? 0 : start;
  const _end = end === undefined ? start : end;

  const output = [];
  let current = _start;
  while (current < _end) {
    output.push(current);
    current += step;
  }
  return output;
}

```

## PR Checklist

### DIFF
--- a/packages/common/utils/src/range.ko.md
+++ b/packages/common/utils/src/range.ko.md
@@ -9,6 +9,7 @@ range(
   // 시작할 숫자
   start: number,
   // 끝날 숫자 (주의: `end` 는 포함되지 않습니다)
+  // @default start
   end?: number,
   // 배열에 포함될 숫자의 간격
   // @default 1
--- a/packages/common/utils/src/range.ko.md
+++ b/packages/common/utils/src/range.ko.md
@@ -9,7 +9,7 @@ range(
   // 시작할 숫자
   start: number,
   // 끝날 숫자 (주의: `end` 는 포함되지 않습니다)
-  end: number,
+  end?: number,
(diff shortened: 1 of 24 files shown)


raon0211 (approved):

packages/common/utils/src/range.ko.md
```diff
@@ -9,7 +9,7 @@ range(
   // 시작할 숫자
   start: number,
   // 끝날 숫자 (주의: `end` 는 포함되지 않습니다)
-  end: number,
+  end?: number,
```
raon0211, line 12: It would be nice to documentate what the default value of `end` is.

ChanhyukPark-Tech: I think that's a good idea!
The default value of the `end` value is not specified on the code and if implicitly undefined, it follows the value of start, so should I correct it as follows?

[fix: @toss/utils declare default value for end parameter for UX](https://github.com/toss/slash/pull/116/commits/f2e8bd69ae1fe86758344ce487723487226ee206)
[f2e8bd6](https://github.com/toss/slash/pull/116/commits/f2e8bd69ae1fe86758344ce487723487226ee206)

raon0211: Yes! Thanks!

raon0211 (approved):

Issue #100, opened 2022-10-29
[Feature]: Add emotion position function types


yoonminsang:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: react/emotion-utils

### **Overview**

I want to like code like this `position({top:10,bottom:8})`
But position function did not support this type.

### **Describe the solution you'd like**

I want to add this code.
```
export function position(coordinates: Coordinates): SerializedStyles;`

  const [top, right, bottom, left] = (() => {
    // position(coordinates)
    if (typeof positionOrTopOrCoordinates === 'object') {
      return [
        positionOrTopOrCoordinates.top,
        positionOrTopOrCoordinates.right,
        positionOrTopOrCoordinates.bottom,
        positionOrTopOrCoordinates.left,
      ];
    }
```

and test code
```
  it('position({top: 0, left: 0})', () => {
    const { getByTestId } = render(<div data-testid="test" css={position({ top: 0, left: 0 })} />);

    const el = getByTestId('test');

    expect(el).toBeInTheDocument();

    expect(el).toHaveStyleRule('top', '0');
    expect(el).not.toHaveStyleRule('bottom', '0');
    expect(el).not.toHaveStyleRule('right', '0');
    expect(el).toHaveStyleRule('left', '0');
  });
```

### **Additional context**

I think toss team may not agree my opinion. So I open an issue, not pr. Thank you

HoseungJang:
You can use `position()` util as you want, by specifying the value of position property:

```tsx
<div css={position('absolute', { top: 10, bottom: 8 })} />
```

Do you feel the same difficulty when you use `position()` like above?

yoonminsang:
In general, but there are exceptions.

I am making Tooltip component. This is my code. I don't want duplicate `absolute`.
But it may be my own thoughts.

```
  const getArrowPositions = useCallback(() => {
    switch (position) {
      case 'top-start':
        return css`
          &::after {
            ${theme.position({ bottom: 0, left: ARROW_WIDTH })}
            margin-bottom: -${ARROW_WIDTH}px;
            ${topBorderColor}
          }
        `;
      case 'top':
        return css`
          &::after {
            ${theme.position({ bottom: 0, left: '50%' })}
            margin-left: -${ARROW_HEIGHT}px;
            margin-bottom: -${ARROW_WIDTH}px;
            ${topBorderColor}
          }
        `;
    // skip
  },[]);
```

```
      <div
        css={css`
          // skip
          ${arrow && [
[...]

PR #117, opened 2022-11-03
feat(emotion-utils): Add position's function method(absolute, fixed, sticky)

yoonminsang:
## Overview

- Add position's function method(absolute, fixed, sticky)
```
position.absolute({ top: 0, left: 0 });
position.absolute(0, 0, 0, 0);
```
- Add example code as comments
- Add test case
- Duplicate code exists. I tried several ways but failed. Please let me know if there is a better way
- cc https://github.com/toss/slash/issues/100

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/utils/position.test.tsx
+++ b/packages/react/emotion-utils/src/utils/position.test.tsx
@@ -47,7 +47,7 @@ describe('position 테스트', () => {
     expect(el).toHaveStyleRule('left', '0');
   });
 
-  describe('position 메서드', () => {
+  describe('position methods', () => {
     describe('absolute', () => {
       it('position.absolute({top: 0, left: 0})', () => {
         const { getByTestId } = render(<div data-testid="test" css={position.absolute({ top: 0, left: 0 })} />);
--- a/packages/react/emotion-utils/src/utils/position.test.tsx
+++ b/packages/react/emotion-utils/src/utils/position.test.tsx
@@ -46,4 +46,96 @@ describe('position 테스트', () => {
     expect(el).not.toHaveStyleRule('right', '0');
     expect(el).toHaveStyleRule('left', '0');
   });
+
+  describe('position 메서드', () => {
+    describe('absolute', () => {
+      it('position.absolute({top: 0, left: 0})', () => {
+        const { getByTestId } = render(<div data-testid="test" css={position.absolute({ top: 0, left: 0 })} />);
+
+        const el = getByTestId('test');
+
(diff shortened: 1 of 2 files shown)


packages/react/emotion-utils/src/utils/position.test.tsx
```diff
@@ -46,4 +46,96 @@ describe('position 테스트', () => {
     expect(el).not.toHaveStyleRule('right', '0');
     expect(el).toHaveStyleRule('left', '0');
   });
+
+  describe('position 메서드', () => {
```
raon0211, line 50: We are planning to migrate all our tests code to English. What about using English here?

```suggestion
  describe('position methods', () => {
```

packages/react/emotion-utils/src/utils/position.ts
```diff
@@ -104,3 +108,45 @@ function isPositionValue(value: unknown): value is Property.Position {
...
+function sticky(coordinates: Coordinates): SerializedStyles;
+function sticky(top: CSSPixelValue, right: CSSPixelValue, bottom: CSSPixelValue, left: CSSPixelValue): SerializedStyles;
+function sticky(topOrCoordinates: Coordinates | CSSPixelValue, ...values: CSSPixelValue[]) {
+  // position(position, coordinates);
+  if (!isCSSPixelValue(topOrCoordinates)) {
+    return position('sticky', topOrCoordinates);
+  }
+  // position(position, top, right, bottom, left);
+  return position('sticky', topOrCoordinates, values[0], values[1], values[2]);
+}
```
HoseungJang, line 152: How about this? @yoonminsang @raon0211 

```typescript
position.absolute = createPosition('absolute');
position.fixed = createPosition('fixed');
position.sticky = createPosition('sticky');

function createPosition(pos: Property.Position) {
  function func(coordinates: Coordinates): SerializedStyles;
  function func(top: CSSPixelValue, right: CSSPixelValue, bottom: CSSPixelValue, left: CSSPixelValue): SerializedStyles;
  function func(topOrCoordinates: Coordinates | CSSPixelValue, ...values: CSSPixelValue[]) {
    // position(position, coordinates);
    if (!isCSSPixelValue(topOrCoordinates)) {
      return position(pos, topOrCoordinates);
    }
    // position(position, top, right, bottom, left);
    return position(pos, topOrCoordinates, values[0], values[1], values[2]);
  }

  return func;
}
```

yoonminsang: That's exactly what I wanted. I think it's good.

PR #118, opened 2022-11-04
feat(a11y): write test code ScreenReaderOnly function

ChanhyukPark-Tech:
## Overview

- set jest environment for @toss/a11y package
- write test code for ScreenReaderOnly function
- replace default export to named export

## PR Checklist

### DIFF
--- a/packages/react/a11y/src/ScreenReaderOnly/ScreenReaderOnly.spec.tsx
+++ b/packages/react/a11y/src/ScreenReaderOnly/ScreenReaderOnly.spec.tsx
@@ -3,8 +3,8 @@ import { ScreenReaderOnly } from '.';
 
 describe('ScreenReaderOnly', () => {
   it('should have proper style', () => {
-    const { getByTestId } = render(<ScreenReaderOnly data-testid="fixture">토스</ScreenReaderOnly>);
-    const fixture = getByTestId('fixture');
+    const { getByText } = render(<ScreenReaderOnly>토스</ScreenReaderOnly>);
+    const fixture = getByText('토스');
 
     expect(fixture).toHaveStyle({
       position: 'absolute',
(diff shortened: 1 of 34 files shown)


packages/react/a11y/src/ScreenReaderOnly/ScreenReaderOnly.spec.tsx
```diff
@@ -0,0 +1,42 @@
+import { render, screen } from '@testing-library/react';
+import { ScreenReaderOnly } from '.';
+
+describe('ScreenReaderOnly', () => {
+  it('should have proper style', () => {
+    const { getByTestId } = render(<ScreenReaderOnly data-testid="fixture">토스</ScreenReaderOnly>);
+    const fixture = getByTestId('fixture');
```
raon0211, line 7: Why not use `screen.getByText('토스')` here, rather than using `getByTestId`? I think we should use testid-based selectors as our last resort.

packages/react/a11y/src/ScreenReaderOnly/ScreenReaderOnly.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+  it('should render children well', () => {
+    render(<ScreenReaderOnly>토스</ScreenReaderOnly>);
+
+    expect(screen.getByText('토스')).toBeInTheDocument();
+    expect(screen.queryByText('toss')).not.toBeInTheDocument();
+  });
+
+  it('should receive other props', () => {
+    const { getByTestId } = render(
+      <ScreenReaderOnly color="white" data-testid="fixture">
```
raon0211, line 34: `ScreenReaderOnly` is not supposed to be given colors, what about giving props like `aria-live` instead?

packages/react/a11y/src/ScreenReaderOnly/ScreenReaderOnly.spec.tsx
ChanhyukPark-Tech: That's right. I knew, but I used a wasted method! I modified it

[fix: do not use test-id it is last resort](https://github.com/toss/slash/pull/118/commits/4f906b0dc5beccd00ba4b14a87993e86db4aacc7)

packages/react/a11y/src/ScreenReaderOnly/ScreenReaderOnly.spec.tsx
ChanhyukPark-Tech: Yes, `ScreenReaderOnly` is not supposed to be given colors at attribute.
I replaced it !
[fix: use more proper attribute](https://github.com/toss/slash/pull/118/commits/2bbcd27b3d751636dee5b487f7a61f46b4ad1c1a)

raon0211 (approved):
Thanks!

PR #125, opened 2022-11-08
feat(rollup-config): Make rollup-config public

artechventure:
## Overview

Make rollup-config public to use outside of this repository.

## PR Checklist

### DIFF
--- a/configs/rollup/package.json
+++ b/configs/rollup/package.json
@@ -1,14 +1,15 @@
 {
   "name": "@slash/rollup-config",
-  "private": true,
   "main": "./src/index.js",
-  "devDependencies": {
+  "dependencies": {
     "@babel/core": "^7.19.3",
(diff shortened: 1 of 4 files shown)


configs/rollup/package.json
```diff
@@ -1,14 +1,15 @@
 {
   "name": "@slash/rollup-config",
```
raon0211, line 2: I think in order for us to make this package public, we have to change the package name. (e.g. `@toss/rollup-config`)

configs/rollup/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@slash/rollup-config",
+  "name": "@toss/rollup-config",
   "main": "./src/index.js",
```
raon0211, line 3: Can you set the initial version here? (Like 1.0.0)

PR #131, opened 2022-11-10
test(react): write test code for useCallbackOnce

ChanhyukPark-Tech:
## Overview

- write test code for `useCallbackOnce`

I checked two concept to preserve author's intend
1. call one time when mounted
2. Even if a component is rerendered, callback is not called more than once

## PR Checklist


raon0211 (approved):
Thanks!

PR #133, opened 2022-11-13
fix(recoil): change directory

manudeli:
## How about change directory for recoil into react from common?
Current position of @toss/recoil, common can make us hard to find it
When I found directory of @toss/recoil, I was confused because of directory

Below picture is description of original [recoil main homepage](https://recoiljs.org/).
> A state management ```library for React```

recoil is library for react
so, How about changing directory of @toss/recoil into react from common

image

## PR Checklist


ChanhyukPark-Tech:
Thank you for giving me a good view!

The library tossteam intended to create with the functions provided by the existing react is located at @toss/react I think.

Even if it's a react only @toss library, it looks like tossteam separating the library, which was created for the easy use of other libraries(framer-motion, recoil).

In addition to the recoil you mentioned, such as framer-motion(@toss/framer-motion) is separated, so I can know the intention of tossteam!

image

ChanhyukPark-Tech:
@toss/framer-motion is in the react folder, so I think it's a good PR!

raon0211 (approved):
Seems `react` directory is more relevant. Thanks @manudeli !

PR #134, opened 2022-11-13
docs(utils): remove back space(%08) in isDifferentArray.spec.ts file name

ChanhyukPark-Tech:
## Overview

image

file name is changed without any change

image

there are two docs `isDifferentArray`

When I looked it up, there was a backspace(%08) hidden in front of the word spec in the file name.

## PR Checklist

- [x] I read and included theses actions below

1. I have read the [Contributing Guide](https://github.com/toss/slash/blob/main/.github/CONTRIBUTING.md)
2. I have written documents and tests, if needed.

PR #138, opened 2022-11-14
fix(utils): change folder isDifferentArray src to src/array

ChanhyukPark-Tech:
## Overview
isDifferentArray is related with array so I changed directory
related with #134  !

## PR Checklist


raon0211 (approved):
Thanks!

PR #143, opened 2022-11-14
feat(context): add all context's displayName only on development

manudeli:
## Add all Context's [displayName](https://reactjs.org/docs/context.html#contextdisplayname) only on development

when using @toss/slash's Contexts on development, I can't find context intuitively in [react-devtools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=ko) because these contexts have no displayName. I just found wrapper function component name (like OverlayProvider have children just Context.Provider, not OverlayContext.Provider)

but after build, I want to remove displayName on react-devtools.

### Reference: [How to add displayName to next/router's RouterContext](https://github.com/vercel/next.js/blob/136b75439612bf6f2f0cd3fd0d8226fdaa8c0f95/packages/next/shared/lib/router-context.ts)
```typescript
import React from 'react'
import { NextRouter } from './router/router'

export const RouterContext = React.createContext<NextRouter>(null as any)

if (process.env.NODE_ENV !== 'production') {
  RouterContext.displayName = 'RouterContext'
}
```

so I added a code line to check process.env.NODE_ENV !== "production" for adding Context.displayName
like next/router's RouterContext

### After `yarn build`, I checked `dist`
image

## Check if new devDependency(@types/node) in @toss/use-overlay is ok, please
I added @types/node as devDependency in @toss/use-overlay
to use process.NODE_ENV, @types/node is required to use process

check it please

## PR Checklist


raon0211 (approved):
This is great. Thanks a lot!

PR #145, opened 2022-11-15
fix(use-overlay): remove unused props

ChanhyukPark-Tech:
## Overview

1. remove unused props `containerId` it is not used
2. change key of array 
The value id is generated uniquely every time overlay is used, so I think it can be used as the key of the array. Is there a any reason for using new Date() as a key?

## PR Checklist


raon0211 (approved):
Thanks!

PR #147, opened 2022-11-16
feat(utils): Add pick function for util object

ChanhyukPark-Tech:
## Overview

Add two functions in utils object

- pick 
  - A utility function to construct new object by picking the key-value pairs corresponding to the given keys from the object.
- split (should be issued) (not included in this pr)
  - Return array two objects The first object consists of picked key values, and the second object omitted the selected key values.

## PR Checklist


packages/common/utils/src/object/split.ts
```diff
@@ -0,0 +1,9 @@
+/** @tossdocs-ignore */
+import { ObjectKeys, omit, pick } from '.';
+
+export function split<ObjectType extends Record<PropertyKey, unknown>, KeyTypes extends Array<ObjectKeys<ObjectType>>>(
+  obj: ObjectType,
+  keys: KeyTypes
+): [ReturnType<typeof pick<ObjectType, KeyTypes>>, ReturnType<typeof omit<ObjectType, KeyTypes>>] {
+  return [pick(obj, keys), omit(obj, keys)];
```
HoseungJang, line 8: This would be more efficient. This iterates over all keys in `obj` only once.

```typescript
const keysSet = new Set(keys);
const picked = {};
const omitted = {};

for (const key in obj) {
  if (keysSet.has(key)) {
    picked[key] = obj[key];
  } else {
    omitted[key] = obj[key];
  }
}

return [picked, omitted];
```

packages/common/utils/src/object/pick.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ArrayUnion, ObjectKeys, objectKeys } from '.';
+
+export function pick<ObjectType extends Record<PropertyKey, unknown>, KeyTypes extends Array<ObjectKeys<ObjectType>>>(
+  obj: ObjectType,
+  keys: KeyTypes
+) {
+  return objectKeys(obj)
+    .filter((k): k is Exclude<ObjectKeys<ObjectType>, ArrayUnion<KeyTypes>> => keys.includes(k))
+    .reduce((acc, key) => ((acc[key] = obj[key]), acc), {} as Pick<ObjectType, ArrayUnion<KeyTypes>>);
+}
```
HoseungJang, line 11: I think it doesn't have to iterate over all keys in `obj`.

This would be more efficient:

```typescript
const picked = {};
for (const key of keys) {
  picked[key] = obj[key];
}
return picked as Pick<ObjectType, ArrayUnion<KeyTypes>>;
```

packages/common/utils/src/object/types.ts
```diff
@@ -0,0 +1,2 @@
+/** @tossdocs-ignore */
+export type ArrayUnion<Type extends readonly unknown[]> = Type[number];
```
HoseungJang, line 2: I think `ArrayElements` is more appropriate name.

packages/common/utils/src/object/split.ts
ChanhyukPark-Tech: thank you!

packages/common/utils/src/object/pick.ts
ChanhyukPark-Tech: image

your code is 80% faster than my code.
I learned it! thank you

ChanhyukPark-Tech: readability is also increased!

ChanhyukPark-Tech: I refactored the[`omit`](https://github.com/toss/slash/blob/main/packages/common/utils/src/object/omit.ts) function similarly, but the performance is rather poor I think it's because the object was copied.

image

ChanhyukPark-Tech:
`ArrayElements `
Is the location of the file appropriate there?
It's a utility type that can be used frequently, so I think it can be defined in [`@toss/utility-types`](https://www.npmjs.com/package/@toss/utility-types), but I'm curious what you think.

HoseungJang (commented):
Thank you for your great contribution.

`pick` looks good but could you exclude `split` in this PR and then open an issue? I want to discuss more about it.

HoseungJang:
> `ArrayElements `
Is the location of the file appropriate there?
It's a utility type that can be used frequently, so I think it can be defined in [`@toss/utility-types`](https://www.npmjs.com/package/@toss/utility-types), but I'm curious what you think.

@ChanhyukPark-Tech Please do this in other PR.

packages/common/utils/src/object/pick.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+describe('pick', () => {
+  it('should work well array of keys is empty', () => {
+    const country = {
+      KR: 'KR',
+      US: 'US',
+      JP: 'JP',
+    } as const;
+
+    expect(pick(country, [])).toStrictEqual({});
+  });
```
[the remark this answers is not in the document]
ChanhyukPark-Tech: thank you [fix: join test case pick and omit function](https://github.com/toss/slash/pull/147/commits/0307472712b43c113cc5fe1913c3bef82445581f)

packages/common/utils/src/object/index.ts
```diff
@@ -3,3 +3,7 @@ export * from './object-entries';
 export * from './object-keys';
 export * from './object-values';
 export * from './omit';
+export * from './pick';
+export * from './split';
+export * from './types';
```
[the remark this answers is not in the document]
ChanhyukPark-Tech: [fix: types do not export in index.ts file](https://github.com/toss/slash/pull/147/commits/9f3bf8729517a8c6cd222c2e3f363b49ef020344)

HoseungJang:
@ChanhyukPark-Tech https://github.com/toss/slash/pull/147#pullrequestreview-1185410290

ChanhyukPark-Tech:
@HoseungJang  I modified [fix: types do not export in index.ts file](https://github.com/toss/slash/pull/147/commits/9f3bf8729517a8c6cd222c2e3f363b49ef020344) is it incorrect?

HoseungJang:
Yes, It looks good.

HoseungJang:
@ChanhyukPark-Tech https://github.com/toss/slash/pull/147#pullrequestreview-1185410290
> pick looks good but could you exclude split in this PR and then open an issue about it? I want to discuss it more.

HoseungJang (approved):
Thanks.

PR #149, opened 2022-11-16
fix(emotion-utils): remove chance to put wrong props in Stack.Horizontal/Vertical, Flex.Center/CenterVertical/CenterHorizontal

manudeli:
# Remove chance to put wrong props in Stack.Horizontal/Vertical, Flex.Center/CenterVertical/CenterHorizontal
### 1. There was chance to put wrong props in them because of wrong compound component type

image

### 2. Not only wrong TypeScript auto complete, Also JavaScript put wrong props in component too.
For example, if Flex.Center accept `align` or `justify`, props will overide origin align, jusify too.
because of props ordering

```ts
Flex.Center = forwardRef<HTMLElement, FlexProps<keyof JSX.IntrinsicElements>>(function Center(props, ref) {
  return <BaseFlex align="center" justify="center" {...props} ref={ref} />;
});

Flex.CenterVertical = forwardRef<HTMLElement, FlexProps<keyof JSX.IntrinsicElements>>(function CenterVertical(
  props,
  ref
) {
  return <BaseFlex align="center" {...props} ref={ref} />;
});

Flex.CenterHorizontal = forwardRef<HTMLElement, FlexProps<keyof JSX.IntrinsicElements>>(function CenterHorizontal(
  props,
  ref
) {
  return <BaseFlex justify="center" {...props} ref={ref} />;
});
```

### 1. Omit wrong property
```ts
type FlexReturnType<O extends string = ''> = <T extends keyof JSX.IntrinsicElements = 'div'>(
  props: Omit<FlexProps<T>, O> & { ref?: Ref<InferenceHTMLElement<T>> }
) => ReactElement | null;

type FlexType = typeof BaseFlex & {
  Center: FlexReturnType<'align' | 'center'>;
  CenterVertical: FlexReturnType<'align'>;
  CenterHorizontal: FlexReturnType<'align'>;
};

export const Flex = BaseFlex as FlexType;

```

### 2. Reorder props injecting
```ts
Flex.Center = forwardRef<HTMLElement, ComponentProps<typeof Flex.Center>>(function Center(props, ref) {
  return <BaseFlex {...props} align="center" justify="center" ref={ref} />;
});

Flex.CenterVertical = forwardRef<HTMLElement, ComponentProps<typeof Flex.CenterVertical>>(function CenterVertical(
  props,
  ref
) {
  return <BaseFlex {...props} align="center" ref={ref} />;
});

Flex.CenterHorizontal = forwardRef<HTMLElement, ComponentProps<typeof Flex.CenterHorizontal>>(function CenterHorizontal(
  props,
  ref
) {
  return <BaseFlex {...props} justify="center" ref={ref} />;
});

```

## Result
1. typescript auto complete will correct someone's mistake.
image

2. also even if javascript developer without typescript autocomplete, they cannot use corner case because of corrected props order

## It can be BREAKING CHANGE to someone
It can be **BREAKING CHANGE** to someone who use wrong case like below picture
I think major version up is required if this change merged.

image

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/Stack.tsx
+++ b/packages/react/emotion-utils/src/Stack.tsx
@@ -1,6 +1,6 @@
 /** @tossdocs-ignore */
 /** @jsxImportSource @emotion/react */
-import { ComponentType, forwardRef, ReactElement, Ref } from 'react';
+import { ComponentProps, forwardRef, ReactElement, Ref } from 'react';
 import { Flex, FlexOptions } from './flex';
 import { gutter, GutterOptions } from './gutter';
 import { AsProps, AxisDirection, InferenceHTMLElement } from './types';
@@ -13,8 +13,8 @@ interface StackProps<T extends keyof JSX.IntrinsicElements = 'div'>
   gutter?: number;
 }
 
-type StackReturnType = <T extends keyof JSX.IntrinsicElements = 'div'>(
-  props: StackProps<T> & { ref?: Ref<InferenceHTMLElement<T>> }
+type StackReturnType<O extends string = ''> = <T extends keyof JSX.IntrinsicElements = 'div'>(
+  props: Omit<StackProps<T>, O> & { ref?: Ref<InferenceHTMLElement<T>> }
 ) => ReactElement | null;
 
 const BaseStack = forwardRef<HTMLElement, StackProps>(function BaseStack(props, ref) {
@@ -32,22 +32,17 @@ const BaseStack = forwardRef<HTMLElement, StackProps>(function BaseStack(props, 
 }) as StackReturnType;
 
 export const Stack = BaseStack as typeof BaseStack & {
-  Vertical: ComponentType<StackWithDirectionProps>;
-  Horizontal: ComponentType<StackWithDirectionProps>;
+  Vertical: StackReturnType<'direction'>;
+  Horizontal: StackReturnType<'direction'>;
 };
 
-type StackWithDirectionProps = Omit<StackProps<keyof JSX.IntrinsicElements>, 'direction'>;
-
-Stack.Horizontal = forwardRef<HTMLElement, StackWithDirectionProps>(function StackHorizontal(props, ref) {
+Stack.Horizontal = forwardRef<HTMLElement, ComponentProps<typeof Stack.Horizontal>>(function StackHorizontal(
+  props,
+  ref
+) {
   return <Stack {...props} direction="horizontal" ref={ref} />;
 });
 
-Stack.Vertical = forwardRef<HTMLElement, StackWithDirectionProps>(function StackVertical(props, ref) {
+Stack.Vertical = forwardRef<HTMLElement, ComponentProps<typeof Stack.Vertical>>(function StackVertical(props, ref) {
   return <Stack {...props} direction="vertical" ref={ref} />;
 });
-
-const Comp = () => {
-  return <Stack.Horizontal direction="horizontal"></Stack.Horizontal>;
-};
-
-console.log(Comp);
--- a/packages/react/emotion-utils/src/Stack.tsx
+++ b/packages/react/emotion-utils/src/Stack.tsx
@@ -1,7 +1,6 @@
 /** @tossdocs-ignore */
 /** @jsxImportSource @emotion/react */
-import { forwardRef, ReactElement, Ref } from 'react';
-
+import { ComponentType, forwardRef, ReactElement, Ref } from 'react';
 import { Flex, FlexOptions } from './flex';
 import { gutter, GutterOptions } from './gutter';
 import { AsProps, AxisDirection, InferenceHTMLElement } from './types';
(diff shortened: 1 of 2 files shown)


HoseungJang (commented):
I think we can keep the current implementation.

Although these components basically center its children, I think there is need to extend css of them.

For example:
```tsx
// There is a FullPage component which vertically centers its children.
function FullPage(props: FlexProps) {
  return <Flex.Center direction='column' css={size.full} {...props} />
}

// But someone wants FullPage to horizontally center its children.
<FullPage direction='row' align='flex-start' />
```

What do you think? @manudeli @raon0211

manudeli:
> I think we can keep the current implementation.
> 
> I think there is need to extend css of these components although they basically center its children.
> 
> For example:
> 
> ```tsx
> // There is a FullPage component which vertically centers its children.
> function FullPage(props: FlexProps) {
>   return <Flex.Center direction='column' css={size.full} {...props} />
> }
> 
> // But someone wants FullPage to horizontally center its children.
> <FullPage direction='row' align='flex-start' />
> ```
> 
> What do you think? @manudeli @raon0211

That's why I submit this Pull Request
I think Flex.Center must be align Center because name of this component is Center
I want to prevent chance that someone override Flex.Center's align because of this name is Center

If Flex.Center's align:center can be overrided, developer can be confused. because Flex.Center name is Center
I think your example's Flex.Center inside of FullPage need to be like below code
```ts
function FullPage(props: FlexProps) {
  return <Flex align='center' direction='column' css={size.full} {...props} /> // change your example to Flex overridable center
}

<FullPage direction='row' align='flex-start' />
```

### 1. Prevent Flex.Center align override
Flex.Center have to return ***Centered Flexed*** element, even though some try overrride
```ts
type FullPageProps = ComponentProps<typeof Flex.Center> // FlexProps is too wide props, because inner Flex.Center:narrowed component is not Flex

function FullPage(props: FullPageProps) {
  return <Flex.Center direction="column" {...props} />; // align center will be not overrided because this component name is Flex.Center
}
<FullPage direction="row" align="flex-start" />;

```

### 2. If someone want to override align inner component
use Flex, not Flex.Center
```ts
type FullPageProps = ComponentProps<typeof Flex> // If you want to override Flex, I think someone need to use Flex too inside. isn't it more clear?

function FullPage(props: FullPageProps) {
  return <Flex direction="column" align="center" {...props} />;
}
<FullPage direction="row" align="flex-start" />;

```

HoseungJang:
> That's why I submit this Pull Request
I think Flex.Center must be align Center because name of this component is Center
I want to prevent chance that someone override Flex.Center's align because of this name is Center

I understand why you think so. But I mean, UI components should be extendable. Please see again the example that I commented. https://github.com/toss/slash/pull/149#pullrequestreview-1184609614

HoseungJang:
I think that:
- the base style is just the base style.
- it should be overridable as needed.
- overriding it would not confuse developers because it's what they intended.

HoseungJang:
When it comes to a component in a design system, I think it can restrict style overriding because it can make the system fragile.

But I think a component such as `Flex.Center` doesn't have to restrict it.

manudeli:
I want to show other reference:@chakra-ui/react I think it's almost same we are discussing
first of all, sorry about comparing @toss/emotion-utils with @chakra-ui/react. I want you not to feel bad about comparing. I just want to suggestion.

@toss/emotion-utils is @toss/emotion-utils. but I think developers can use this library in simmilar reason.

```ts
import { Box, Stack, VStack } from "@chakra-ui/react"

const Comp = () => {
  return (
    <>
      <Stack direction="row">
        <RedBox />
        <RedBox />
        <RedBox />
        <RedBox />
        <RedBox />
        <RedBox />
      </Stack>
      <VStack direction="row">
        <RedBox />
        <RedBox />
        <RedBox />
        <RedBox />
        <RedBox />
        <RedBox />
      </VStack>
    </>
  )
}

export default Comp

const RedBox = () => <Box w="50px" h="50px" bgColor="red" />

```

image

VStack in @chakra-ui/react direction cannot be overridable not like Stack.
I think too wide extendability can make error or confusing implementation not on purpose.

I think That's why use VStack. and Flex.Center too. even though there is props

in addition, I understood your meaning exactly too. Thanks for your replying! 👍 so grateful!

manudeli:
@HoseungJang on the other hand,

I want to ask whether you accept this PR's change about Stack.Vertical/Horizontal without Flex.Center/CenterHorizontal/CenterVertical

HoseungJang:
Thank you for your good example. I think that's exactly what 'confusing' is, because `Vstack` allows `direction` prop but it will be ignored actually.

But our components are not ignoring any passed props so I think that's not 'confusing', that's 'overriding'.

HoseungJang:
> I want to ask whether you accept this PR's change about @toss/emotion-utils's Stack.Vertical/Horizontal without Flex.Center/CenterHorizontal/CenterVertical

My opinion about flex components is the same for stack components. I want to see what other maintainers think about it and then determine what we are going to do.

@raon0211 @changyoungoh

raon0211:
Yes, I agree with @HoseungJang because I see `<Flex.Center />` as just an alias to `<Flex align="center" justify="center" />`. It only provides default values and values can be overridden. For example, even if we don't allow align and justify props to be passed, it can still be overridden by `css` or `className` prop. 

Aside from that I rather want to keep the implementation simple for maintainability if there is no chance of confusion. I think the current implementation is much simpler and easier to maintain.

HoseungJang:
Also, this is a breaking change because someone who are overriding our base styles could encounter an UI-level error.
Adding new features is available anytime, but removing existing features is what we always have to be careful.

manudeli:
I perfectly understood why you don't want to do like this PR👍👍👍
I will close this PR, Thanks for your kind explaining! So grateful!

Issue #146, opened 2022-11-15
[Feature]: How about return exit function in `useOverlay` hook


ChanhyukPark-Tech:
### **Package Scope**

- [x] Add to an existing package

Package name: @toss/use-overlay

### **Overview**

Currently, if I am using `useOverlay`, I can control isOpen to false through the `close` function, but if I want to unmount the overlay using return value of the `useOverlay`, the interface does not exist. how about make exit function to developer use that function to unmount overlay?

## AS-IS
```tsx
 return useMemo(
    () => ({
      open: (overlayElement: CreateOverlayElement) => {
        mount(
          id,
          <OverlayController
            key={id}
            ref={overlayRef}
            overlayElement={overlayElement}
            onExit={() => {
              unmount(id);
            }}
          />
        );
      },
      close: () => {
        overlayRef.current?.close();
      }
    }),
    [id, mount, unmount]
  );
```

## TO-BE

```tsx
 return useMemo(
    () => ({
      open: (overlayElement: CreateOverlayElement) => {
        mount(
          id,
          <OverlayController
            key={id}
            ref={overlayRef}
            overlayElement={overlayElement}
            onExit={() => {
              unmount(id);
            }}
          />
        );
      },
      close: () => {
        overlayRef.current?.close();
      },
      exit: () => {
        unmount(id);
      },
    }),
    [id, mount, unmount]
  );
```

## Usage

```tsx
function App() {
  const overlay = useOverlay();
  
  const openFooConfirmDialog = () => {
    overlay.open(({ isOpen, close, exit }) => (
      <FooConfirmDialog
        open={isOpen}
        onClose={() => {
          close();
        }}
        onConfirm={() => {
          exit();
        }}
      />
    ));
  };

  return (
    <div>
      <button onClick={overlay.close}>just set isOpen to false</button>
      <button onClick={overlay.exit}>unmount</button> // you can use like this to unmount overlay
    </div>
  );
}
```

raon0211:
Seems right. Can you open a pull request?

ChanhyukPark-Tech:
yes! thank you😄

PR #150, opened 2022-11-17
feat(use-overlay): add exit function 

ChanhyukPark-Tech:
## Overview

related with #146

- Abstracts and returns the function so that developers using `useOverlay` can unmount the overlay from the outside.
- wrote proper test code for verify this function

## PR Checklist

### DIFF
--- a/packages/react/use-overlay/src/tests/exit.spec.tsx
+++ b/packages/react/use-overlay/src/tests/exit.spec.tsx
@@ -1,42 +1,35 @@
-import { act, screen, waitFor } from '@testing-library/react';
-import React, { useEffect } from 'react';
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { useOverlay } from '../useOverlay';
 import { renderWithContext } from './utils';
-
-const closeDuration = 1000;
 
 function TestComponent() {
   const overlay = useOverlay();
 
-  useEffect(() => {
-    overlay.open(() => {
-      return <div>toss</div>;
-    });
-
-    const timeoutID = setTimeout(() => {
-      overlay.exit();
-    }, closeDuration);
-
-    return () => clearTimeout(timeoutID);
-  }, [overlay]);
-
-  return null;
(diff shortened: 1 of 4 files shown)


packages/react/use-overlay/src/tests/exit.spec.tsx
```diff
@@ -0,0 +1,42 @@
+import { act, screen, waitFor } from '@testing-library/react';
+import React, { useEffect } from 'react';
+import { useOverlay } from '../useOverlay';
+import { renderWithContext } from './utils';
+
+const closeDuration = 1000;
+
+function OverlayCallerComponent() {
```
raon0211, line 8:
```suggestion
function TestComponent() {
```

We usually name components like these `TestComponent`s. What about following the convention?

packages/react/use-overlay/src/tests/exit.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    const timeoutID = setTimeout(() => {
+      overlay.exit();
+    }, closeDuration);
+
+    return () => clearTimeout(timeoutID);
+  }, [overlay]);
+
+  return null;
```
raon0211, line 23: What about adding buttons like `open`, and `exit` here, which opens the overlay and unmounts it, and writing the code as below?

```tsx
const openButton = screen.getByText('open');
const exitButton = screen.getByText('exit');

await userEvent.click(openButton);

expect(screen.getByText('toss')).toBeInTheDocument();

await userEvent.click(closeButton);

expect(screen.queryByText('toss')).not.toBeInTheDocument();

raon0211: This test code will not require `jest.advanceTimersByTime()` so I think it's much simpler.

ChanhyukPark-Tech: Thx I made it!

raon0211 (approved):

PR #152, opened 2022-11-19
feat(utility-types): arrayElements type 

ChanhyukPark-Tech:
## Overview

When I use typescript, I may need to obtain type values for items in the array. 
So  added it to the package.

## PR Checklist


packages/common/utils/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.14.8",
+    "@toss/utility-types": "workspace:^1.0.5",
```
raon0211, line 22: This adds an additional dependency to our `utils` library (and even to users who do not use TypeScript), so I am a little bit concerned about if it is worth it. Why not just adding `arrayElements` to `utility-types`, rather than refactoring some of our code?

packages/common/utility-types/src/arrayElements.en.md
```diff
@@ -0,0 +1,3 @@
+# arrayElements
```
raon0211, line 1: According to [other TypeScript built-in types](https://www.typescriptlang.org/docs/handbook/utility-types.html) I guess names like `ElementType<>` follows the convention. What do you think?

packages/common/utils/package.json
ChanhyukPark-Tech: Thank you for your kind words.
It could be a waste for people who don't use typescript!

packages/common/utility-types/src/arrayElements.en.md
ChanhyukPark-Tech: Yes I think type or interface should be Pascal case😁

ChanhyukPark-Tech: The name ElementType is more intuitive

raon0211 (approved):

raon0211:
Thanks, @ChanhyukPark-Tech !

PR #154, opened 2022-11-22
feat(utils): Allow generic argument in `get`

ChanhyukPark-Tech:
## Overview

The type of `get` method is `any`.
Values such as `null` cannot be returned, which can lead to misunderstandings.

## PR Checklist


packages/common/utils/src/get-set_get.en.md
```diff
@@ -11,7 +11,7 @@ function get(
   path: string,
   // @default undefined
   defaultValue?: any
-): any;
+): Record<string, any> | undefined;
```
raon0211, line 14: The type is actually `any` because `obj`'s type is any 😢 

We might write a smart type using [template literal types](https://toss.tech/article/template-literal-types), but currently the type `any` is right. (It is the actual implementation)

ChanhyukPark-Tech: Isn't the type of `obj` not `any` but `Record<string, any>` ? get method cannot receive a value such as `null`.

image

raon0211: Oh, you are in fact right. However since the value type of `obj` is any, the return type would be any.

raon0211: Yes, I think using generics would be better

ChanhyukPark-Tech: I'm sorry for the late reply!
**HoseungJang** gave me a great idea, so I'll change it in that direction

ChanhyukPark-Tech: The current `defaultValue` value which is third parameter is fixed as undefined type, so it is not `any` type as in docs. 

`In docs`

```typescript
function get(
  // 값을 가져올 객체
  obj: Record<string, any>,
  // 가져올 값의 경로
  path: string,
  // 경로에 해당하는 값이 없을 때 반환할 기본값
  // @default undefined
  defaultValue?: any //  <-- this is not in real implementation
): Record<string, any> | undefined;
```
Also there's no test code
If I change the `defaultValue` type to `any` type and abstract it so that the user can enter any value, the return value of the `get` function will be `any` again. 

This is because if the find value is not found, the `defaultValue` is returned as is.
Should I use two generics?
I think it would be better to just return `any` like the present implementation.

HoseungJang: 1. The Return type is not only `Record<string, any>`. The value of object's property can be anything like `number`, `string`, `undefined` and etc. So it should be `any`.
    ```typescript
    const obj = {
      a: {
        b: {
          c: 1234,
        },
      },
      d: "str",
    };
  
    get(obj, "a.b"); // { c: 1234 }
    get(obj, "a.b.c"); // 1234
    get(obj, "d"); // "str"
    ```

2. `defaultValue` parameter isn't fixed as `undefined`, please see [Default Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).

3. I think the return type should be equal to the type of `defaultValue` so you can use only one generic.
    ```typescript
    export function get<T = any>(obj: Record<string, any>, path: string, defaultValue?: T): T {
      const travel = (regexp: RegExp) =>
        String.prototype.split
          .call(path, regexp)
          .filter(Boolean)
          .reduce((acc, key) => (acc !== null && acc !== undefined ? acc[key] : acc), obj);
      const result = travel(/[,[\]]+?/) || travel(/[,[\].]+?/);
      return (result === undefined || result === obj ? defaultValue : result) as T;
    }
    ```

ChanhyukPark-Tech: Thank you for a good example!😁

Wouldn't using only one generic cause the following human error?

```tsx
const obj = {
  a: {
    b: {
      c: 1234,
    },
  },
  d: 'str',
};

const expectedString =  get(obj, 'a.b.c', '1234');

console.log(expectedString); // this is not "1234" but 1234 this can make runtime error
```

image


[this is my playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBA5gU1gXhgHgCo1QQzATwD4AKEAIwCsAuGAJQVACcATNaRgSzDgBoY8ifAA44oACxrsuvGMwQAzHAFcANlABqOFUoQB+GhgCUBmAB8YSsHPlcEzbIRgBvAFAwYoSLCiMcANwQVbBhiRgREAA8hGno4AFEowwc3dxgAZR9pADohRhAofPwhBCyIIRUOKBTUmCzgLRViEXE+MMihQ2rUrJs1BEZiACEQEBUEPE6a9yyw5iVgBGJiHGBgPgBrBHwk5Edl1ZgAQmRUMFUggDIL-gPj1EtrW3tdG+AAbU38AF0YGhXgQx8cgUQwAbhSnmgMDCEFUKBgPn8gWIAHo3jw3gAdL5fADUuhRSVM5kRAUaaIx2KyeIJYJSYSgSkYYBCMLh2BOFisCieZnMbLUHNQwJgL2syjUmm0CF+0IQsLUSRwEBgGHBAF9wS5IbARahXO4cDQDakyMauu5gDQAIwAJgAzAAWHhddUu9xulLMGgAcnYPpdmpc2vAUIQUQYUDsGU43GC8CQpEofB9OCyZDqAZgPrtTp9dODIcgoxKKhAcGI4eKwCjzBj0jBMBRKIRYg4KvbMDA+RgACJc47ezAyEpYAPW536iyALY4TbQyxQDjTmX9PKMIA)

thank you!

raon0211: Yes this may bring an error, but we trust the developer to provide the correct type. For example [ky](https://github.com/sindresorhus/ky) has a similar approach.

```tsx
// foo.com/number may not return a string
const result = await ky('https://foo.com/number').json<string>();
```

ChanhyukPark-Tech: Thank you for the good example!
Even if it was a simple function, it was an opportunity to think a lot from the perspective of the developer who actually uses it get function.

raon0211 (approved):
Thanks1

PR #155, opened 2022-11-22
fix(emotion-utils): change name of  `StackWithDirectionProps` props

ChanhyukPark-Tech:
## Overview

I changed name of props `StackWithDirectionProps` => `StackWithoutDirectionProps`

Since the role of props is the type that remove `direction` from the basic `StackProps`, wouldn't `StackWithoutDirectionProps` be more appropriate for the type name than `StackWithDirectionProps`?

I'm curious about your opinion!

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/Stack.tsx
+++ b/packages/react/emotion-utils/src/Stack.tsx
@@ -39,12 +39,12 @@ type StackType = typeof BaseStack & {
 
 export const Stack = BaseStack as StackType;
 
-type StackWithDirectionProps = Omit<StackProps<keyof JSX.IntrinsicElements>, 'direction'>;
+type StackWithoutDirectionProps = Omit<StackProps<keyof JSX.IntrinsicElements>, 'direction'>;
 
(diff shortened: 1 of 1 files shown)


manudeli (commented):
In my opinion, StackWithDirectionProps is more proper.

Because That type is scoping Stack component self, not props.

If you want to expose meaning of omit, I think StackPropsWithoutDirection is more proper naiming.

# Scope difference
- 'StackProps'WithoutDirection (Scoping Omit type)
- 'StackWithDirection'Props (Scoping Component self)

Then, I think Stack.Vertical is StackWithDirection. So StackWithDirectionProps is proper for naming Props for Stack.Vertical or others

ChanhyukPark-Tech:
@manudeli  Thank you for the review!
I've been reading for quite a while and finally I understand.

`StackWithDirectionProps` has the meaning of Props, which is **already** used for Stack with `directionality`.

I saw the naming from a very new perspective. If I want to do it my way, `StackPropsWithoutDirection` is the right way to do it, but I thought it would be better to keep it the original implementation after seeing the review!

PR #159, opened 2022-11-30
chore: add dist, esm to ignore eslint after all packages build

manudeli:
## Overview

add dist esm in .eslintignore to ignore eslint after all packages build
I don't want to see error in dist esm, and don't want to let them can be formatted on save.

image

## PR Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+dist
+esm


raon0211 (approved):
Thanks!

Issue #163, opened 2022-12-05
[Feature]: Why needed two RegExp in the get function?


leokang94:
### **Package Scope**

- [X] Add to an existing package
- [ ] New package

Package name: @toss/utils

### **Overview**

current `get` function uses about two RegExp(`/[,[\]]+?/`, `/[,[\].]+?/`).

```ts
export const get = <T = any>(obj: Record<string, any>, path: string, defaultValue?: T): T => {
  // ...
  const result = travel(/[,[\]]+?/) || travel(/[,[\].]+?/);
  // ...
};
```

### **Describe the solution you'd like**

I think the second RegExp can sufficiently replace the first RegExp like below.
```ts
export const get = <T = any>(obj: Record<string, any>, path: string, defaultValue?: T): T => {
  // ...
  const result = travel(/[,[\].]+?/);
  // ...
};
```

### **Additional context**

If you need it because of another edge case that I don't know, could you let me know?

raon0211:
Yes, I think we could definitely simplify the function implementation. Do you have some time to open a pull request?

leokang94:
Of course, Thanks!
I created PR #164 😀

PR #164, opened 2022-12-05
fix(utils): remove unnecessary logic in get

leokang94:
## Overview

ref #163
### AS-IS
The current `get` function has `travel` logic with use two RegExp as below.

```ts
export const get = <T = any>(obj: Record<string, any>, path: string, defaultValue?: T): T => {
  // ...
  const result = travel(/[,[\]]+?/) || travel(/[,[\].]+?/);
  // ...
};
```

However, the first RegExp is sufficient to replace the second RegExp, so the first is unnecessary.

### TO-BE
removed first RegExp as below.

```ts
export const get = <T = any>(obj: Record<string, any>, path: string, defaultValue?: T): T => {
  // ...
  const result = travel(/[,[\].]+?/);
  // ...
};
```

## PR Checklist

### DIFF
--- a/packages/common/utils/src/get-set.ts
+++ b/packages/common/utils/src/get-set.ts
@@ -5,7 +5,7 @@ export const get = <T = any>(obj: Record<string, any>, path: string, defaultValu
       .call(path, regexp)
       .filter(Boolean)
       .reduce((acc, key) => (acc !== null && acc !== undefined ? acc[key] : acc), obj);
-  const result = travel(/[,[\]]+?/) || travel(/[,[\].]+?/);
+  const result = travel(/[,[\].]+?/);
   return (result === undefined || result === obj ? defaultValue : result) as T;
 };
 


raon0211:
Last night I thought about this issue and actually found out why there are two regular expressions. 

Consider the following object:

```
{
  "foo.bar": 2,
  "foo": {
    "bar": 1
  },
}
```

In the current implementation, `get(obj, "foo.bar")` would return `2`, not `1`. If we change the implementation, it would return `1`. 

I don't know whether this is an intended behavior...

leokang94:
Oh, I see.
Umm.. If that above case could be return `2` as you told me, I think it should be return `1` at the follow case.

image
But.. I don't know that is right case.. 🤔

raon0211:
Yes, that's right, and I think we might use three RegExps. However I think this is quite a minor case and not sure about whether or not we should support this behavior...

HoseungJang:
I think we don't have to support keys including separator like `"a.b.c"` because there are too many available separators(`.`, `[]`, `,`). If there were only one separator we would support it.

raon0211:
Thanks, @leokang94 !

PR #165, opened 2022-12-06
refactor(package.json): remove unnecessary files of all packages

manudeli:
ISSUE CLOSE: #166 

## Narrow only packages's necessary files by using files field of package.json
If not using files field, @toss/slash user can see unnecessary files in node_modules like @toss/error-boundary's jest.config.js, jest.setup.ts etc.
image

### [files field of package.json](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files)

I used files filed as ["dist", "esm"], but we don't have to worry about ignoring necessary files like package.json, README, LICENSE that almost be used as filename in @toss/slash. because below filename will be contained in package defaultly.

Certain files are always included, regardless of settings:
- package.json
- README
- LICENSE / LICENCE
- The file in the "main" field
- README & LICENSE can have any case and extension.

package.json have to narrow files to publish package contain only necessary. by using files field, we can narrow package's files. so I added files field in 05e1c07

### Only @toss/ky, I didn't add this solution in 05e1c07, help needed.
@toss/ky have other directory structure that should be published is different with other packages: files: ["dist", "esm"].

**Could you help this PR to narrow @toss/ky's package after build?**

## PR Checklist

### DIFF
--- a/packages/common/assert/package.json
+++ b/packages/common/assert/package.json
@@ -10,6 +10,10 @@
     "./package.json": "./package.json"
   },
   "main": "src/index.ts",
+  "files": [
+    "dist",
+    "esm"
+  ],
   "scripts": {
     "build": "rm -rf dist esm && tsc --declaration --emitDeclarationOnly --declarationDir dist && rollup -c rollup.config.js",
     "prepack": "yarn build",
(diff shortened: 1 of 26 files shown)


raon0211:
We will check `@toss/ky` after this pull request. Thanks, @manudeli !

raon0211 (approved):

PR #167, opened 2022-12-07
docs(scroll-restoration): Remove empty docs

saseungmin:
## Overview

The title of this document is "scrollRestoration.set". so, unnecessary menu has been created in the sidebar.
In addition nothing to see in the "scrollRestoration" [document](https://slash.page/ko/libraries/common/utils/src/scrollRestoration.ts.tossdocs).

url: https://slash.page/ko/libraries/common/utils/src/scrollrestoration.i18n/

스크린샷 2022-12-07 오후 9 50 51

## PR Checklist

### DIFF
--- a/packages/common/utils/src/scrollRestoration.en.md
+++ b/packages/common/utils/src/scrollRestoration.en.md
@@ -1,4 +1,4 @@
-# scrollRestoration
+# scrollRestoration.set
 
 This is a helper that helps you easily set/clear related to scrollRestoration.
 
--- a/packages/common/utils/src/scrollRestoration.ko.md
+++ b/packages/common/utils/src/scrollRestoration.ko.md
@@ -1,4 +1,4 @@
-# scrollRestoration
+# scrollRestoration.set
 
 scrollRestoration 관련 set/clear를 쉽게 할 수 있도록 도와주는 helper입니다.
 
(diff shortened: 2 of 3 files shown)


packages/common/utils/src/scrollRestoration.en.md
```diff
@@ -1,4 +1,4 @@
-# scrollRestoration.set
```
HoseungJang, line 1: 1. I think this is correct title because there is only `set` method in the `scrollRestoration` object, and I'm writing docs per method.
2. A reason why an unnecessary menu named `scrollRestoration` is created is there's no `@tossdocs-ignore` comment on the top of [scrollRestoration.ts](https://github.com/toss/slash/blob/d883a0b2aebdbc2ca39c67902cec754c63921dfe/packages/common/utils/src/scrollRestoration.ts#L1) file.

So could you keep the current title and add `@tossdocs-ignore` comment on the top of the file?

```typescript
/** @tossdocs-ignore */
```

saseungmin: @HoseungJang 
Thank you for the detailed explanation. 🙇🏻‍♂️
The title has been restored, and `@tossdocs-ignore` has been added to the top of the `scrollRestoration.ts`.

raon0211 (approved):

PR #168, opened 2022-12-07
feat(utils): add number type to createKey result value

2-chanhee:
## Overview

Add number type to groupBy function's createKey's return value type
Reduce function's initial value is object, object only allow key of string type. 
So we don't need to add number type to  `result: Record<string, T[]>`
It is useful for handling database results.

### **AS-IS**
```
export function groupBy<T>(data: T[], createKey: (item: T) => string) {
  return data.reduce((result: Record<string, T[]>, current) => {
    const key = createKey(current);
    const value = result[key];
    if (value == null) {
      result[key] = [current];
    } else {
      value.push(current);
    }
    return result;
  }, {});
}
```

### **TO-BE**
```
export function groupBy<T>(data: T[], createKey: (item: T) => string | number) {
  return data.reduce((result: Record<string, T[]>, current) => {
    const key = createKey(current);
    const value = result[key];
    if (value == null) {
      result[key] = [current];
    } else {
      value.push(current);
    }
    return result;
  }, {});
}
```

## PR Checklist

- [✓] I read and included theses actions below

### DIFF
--- a/packages/common/utils/src/groupBy.ts
+++ b/packages/common/utils/src/groupBy.ts
@@ -1,5 +1,5 @@
 /** @tossdocs-ignore */
-export function groupBy<T>(data: T[], createKey: (item: T) => string) {
+export function groupBy<T>(data: T[], createKey: (item: T) => string | number) {
   return data.reduce((result: Record<string, T[]>, current) => {
     const key = createKey(current);
     const value = result[key];
--- a/packages/common/utils/src/groupBy.spec.ts
+++ b/packages/common/utils/src/groupBy.spec.ts
@@ -1,7 +1,7 @@
 import { groupBy } from './groupBy';
 
 describe('groupBy 함수는', () => {
-  const input = [
+  const stringInput = [
     { groupName: '부엉이', value: 1 },
     { groupName: '다람쥐', value: 2 },
     { groupName: '너구리', value: 3 },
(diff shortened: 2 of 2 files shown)


packages/common/utils/src/groupBy.ts
```diff
@@ -1,5 +1,5 @@
 /** @tossdocs-ignore */
-export function groupBy<T>(data: T[], createKey: (item: T) => string) {
+export function groupBy<T>(data: T[], createKey: (item: T) => string | number) {
   return data.reduce((result: Record<string, T[]>, current) => {
```
HoseungJang, line 3: > Reduce function's initial value is object, object only allow key of string type.
So we don't need to add number type to result: Record<string, T[]>

You can define an object with number keys and it can be accessed by both string and number.

```typescript
const obj = { 1: 'foo' };

console.log(obj[1]); // 'foo'
console.log(obj['1']); // 'foo'
```

raon0211: You are right, but I think the below code is not very hard to write, and we would like to keep the implementation as simple as possible, to keep the code maintainable.

```tsx
groupBy(items, item => item.id.toString())
```

Also, numbers are not the only type we should consider. We might have to support boolean as well, if we want to group them by true or false. These decisions might make the code complex, so we want to keep the code as simple as possible, and provide the very basic functionality. (Users of code might wrap our implementation)

2-chanhee: @HoseungJang @raon0211 
I understand your opinion well. Thanks for the detail comments!

PR #178, opened 2023-01-02
fix(error-boundary): connect RenderFallbackProps.reset's args to onReset correctly

manudeli:
## Connect RenderFallbackProps.reset's args to onReset correctly
If we want to pass any args to onReset's args when using RenderFallbackProps.reset, I think merging this Pull Request is required.
For easy understanding why I connect args, I left [the code for reference (bvaughn/react-error-boundary)](https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L78-L81).

### If you merge this Pull Request, below implementation will be available like [bvaughn/react-error-boundary](https://github.com/bvaughn/react-error-boundary)
```tsx
<ErrorBoundary
  renderFallback={({ reset }) => <button onClick={() => reset('something')}>reset all</button>}
  onReset={(...args) => {
    // args[0] will be 'something'
    console.log(args[0])
  }}
>
...
</ErrorBoundary>
```

### Suggestion other way to solve this problem
I think removing this feature will be also good option. differently with bvaughn/react-error-boundary, I think passing args is not essential feature in ErrorBoundary and this feature can make difficulty to understand what is args, so, after discussing, I want to suggest Removing all args in ErrorBoundary. like [@suspensive/react](https://github.com/suspensive/react/blob/main/packages/react/src/ErrorBoundary.tsx#L25)

I want to make ErrorBoundary as simply understand without passing args

thanks to @jinn2u, why I made this PR

## PR Checklist


raon0211:
Our response is slow due to our company’s holiday season, and every maintainer is at vacation right now. We will review this pull request by next week.

raon0211:
I guess this is quite a minor feature -- why not hold this until there are actual use cases? I understand this feature is good-to-have, but it is hard to type the props (we have to use `unknown[]`) and will make our codebase harder to maintain (because we have to support additional feature)

raon0211:
If you agree, can we close this pull request and open this again when we have an actual use case?

manudeli:
> I guess this is quite a minor feature -- why not hold this until there are actual use cases? I understand this feature is good-to-have, but it is hard to type the props (we have to use `unknown[]`) and will make our codebase harder to maintain (because we have to support additional feature)

You mean, you want to remove this feature like my suggestion, right?

Screenshot_20230113_000804_Chrome

I think so too. So If you accept my suggestion: removing not essential feature(passing args: unknown[]), not like [bvaughn/react-error-boundary](https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L78-L81). I will remove args type of reset with this feature. Could you decide it for me?

raon0211:
Yes, I think removing the feature would be nice.

manudeli:
Cool, I removed reset's args type in [01cb9d5](https://github.com/toss/slash/pull/178/commits/01cb9d5bd719fe24a99acc15c3502200321e1bd2) like @suspensive/react's ErrorBoundary's reset

Thanks to @jinn2u fixed it in @suspensive/react

raon0211 (approved):

Issue #184, opened 2023-01-04
[BUG]: clamp returns wrong value


FFloou:
### **Package Scope** 

Package name: @toss/utils

### **Describe the bug**

In docs example, `clamp(3, 5)` returns `5`, but the actual return value is `3`.
If this is not a bug, please change the docs not to be confused!

### **Expected behavior**

Return value of `clamp(3, 5)` to be 5  

### **Possible Solution** 

Change clamp.ts to
```
 if (bound2 == null) {
    return Math.max(value, bound1);
  }
```

PR #185, opened 2023-01-04
fix(utils): Fix clamp return wrong value when value is smaller than min value

alstn2468:
## Overview

PR on issue #184

If the value of `bound2` does not exist, the `clamp` should return `bound1` if value is less than `bound1`.
So I fixed when `bound2` is `null` and modify the test case.

```typescript
function clamp(value: number, min: number): number;
function clamp(value: number, min: number, max: number): number;
```

According to the docs, `bound1` corresponds to min value.
The existing code seems to be written assuming that `bound1` is max value.

## PR Checklist

### DIFF
--- a/packages/common/utils/src/clamp.en.md
+++ b/packages/common/utils/src/clamp.en.md
@@ -9,7 +9,7 @@ Set a lower and upper bound to a value. If the value is smaller than its lower b
 - Caution: When the lower bound is bigger than the upper bound, the `clamp` function throws an error.
 
 ```typescript
-function clamp(value: number, min: number): number;
+function clamp(value: number, max: number): number;
 function clamp(value: number, min: number, max: number): number;
 ```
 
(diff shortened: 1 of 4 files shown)


HoseungJang:
I think `clamp` was designed to be used like:

```typescript
function clamp(value: number, maximumValue: number);
function clamp(value: number, bound1: number, bound2: number);
```

So I think the current implementation is correct and we should keep it.

alstn2468:
Then I'll fix docs and commit it. 👍

raon0211:
Yes I think fixing the docs would be sufficient. 

Although I think it is very hard for us to predict what `clamp(3, 5)`  does — we would always have to check the docs to know if `5` is min or max. 

Better for us to deprecate the API with two arguments in the next release…

raon0211 (approved):

PR #186, opened 2023-01-05
fix(react-query): remove chance to use generic of type SuspendedUseQueryResultOnIdle

manudeli:
## Remove chance to use generic of SuspendedUseQueryResultOnIdle

There was exported type `SuspendedUseQueryResultOnIdle` with generic to type TData. but SuspendedUseQueryResultOnIdle's purpose is for idle status of Promise, so It cannot be filled TData without undefined. so I remove chance to use generic of SuspendedUseQueryResultOnIdle.

## PR Checklist


artechventure (approved):
LGTM

PR #196, opened 2023-01-24
fix(error-boundary): fix useErrorBoundary's return type to object that can have bigger scalability than before

manudeli:
BREAKING CHANGE: return type is different. this will make breaking change.

## fix useErrorBoundary's return type to object that can have bigger scalability than before
I expect useErrorBoundary will provide more feature than now.
but current return type of useErrorBoundary cannot have scalability. cause it's just function.

I think this hook's return type need to be object type including throw field for scalability.

### AS-IS
```tsx
const throwError = useErrorBoundary()

return <button onClick={() => throwError(new Error(....))}/>
```

### TO-BE
```tsx
const errorBoundary = useErrorBoundary()

return <button onClick={() => errorBoundary.throw(new Error(....))}/>
```

### AS-IS
```tsx
const throwError = useErrorBoundary()

return <button onClick={() => throwError(new Error(....))}/>
```

### TO-BE
```tsx
const handleError = useErrorHandler()

return <button onClick={() => handleError(new Error(....))}/>
```

## PR Checklist

### DIFF
--- a/packages/react/error-boundary/src/useErrorBoundary.test.tsx
+++ b/packages/react/error-boundary/src/useErrorBoundary.test.tsx
@@ -24,7 +24,9 @@ describe('useErrorBoundary', () => {
     });
 
     act(() => {
-      result.current(error);
+      const boundary = result.current;
+
+      boundary.throw(error);
     });
 
     // Then
--- a/packages/react/error-boundary/src/useErrorBoundary.ts
+++ b/packages/react/error-boundary/src/useErrorBoundary.ts
@@ -1,5 +1,5 @@
 /** @tossdocs-ignore */
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 
 export default function useErrorBoundary<ErrorType extends Error>() {
   const [error, setError] = useState<ErrorType | null>(null);
@@ -8,5 +8,7 @@ export default function useErrorBoundary<ErrorType extends Error>() {
     throw error;
   }
 
-  return setError;
+  const errorBoundary = useMemo(() => ({ throw: setError }), [setError]);
+
+  return errorBoundary;
 }


raon0211:
Which feature do you expect to provide? I wish this change be done when actual use cases are apparent.

manudeli:
> Which feature do you expect to provide? I wish this change be done when actual use cases are apparent.

Regardless of the addition of new features into useErrorBoundary, I thought it was necessary to maintain consistency with useErrorBoundaryGroup (return group is object but have just only one key(reset))

```tsx
const group = useErrorBoundaryGroup() // if I use use[Component], I expect object returned for Component

return <button onClick={group.reset}/>
```

so, If @toss/slash useErrorBoundary need to return function, how about renaming this hook to useErrorHandler like bvaughn/react-error-boundary. and then make useErrorBoundary wrap useErrorHandler and return object containing return function of useErrorHandler?

## Other solution: rename useErrorBoundary of @toss/error-boundary to useErrorHandler of bvaughn/react-error-boundary

### AS-IS
```tsx
const throwError = useErrorBoundary()

return <button onClick={() => throwError(new Error(....))}/>
```

### TO-BE
```tsx
const throwError = useErrorHandler()

return <button onClick={() => throwError(new Error(....))}/>
```

```tsx
const errorBoundary = useErrorBoundary()

return <button onClick={() => errorBoundary.throw(new Error(....))}/>
```

HoseungJang:
@manudeli I agree with your comment about consistency and confusing of naming, but I think it is currently inappropriate to merge this change.

Because this is a breaking change. I think the scalability you mentioned should be considered at a draft level. But currently `useErrorBoundary` is at a maintaining level.

So we need to approach in something different way to change it now. As @raon0211 [commented](https://github.com/toss/slash/pull/196#issuecomment-1402040307), I think currently we should not make a breaking change without necessary, and apparent needs or use cases, like your changes.

So I think we will close this PR. Any idea?

manudeli:
> @manudeli I agree with [your comment](https://github.com/toss/slash/pull/196#issuecomment-1402163796) about maintaining consistency and confusing of naming, but I think it is currently inappropriate to merge this change.
> 
> Because this is a breaking change. I think the scalability and the consistency that you commented should be considered at a draft level. I think currently `useErrorBoundary` is at a maintaining level. So I'm sorry but we need to approach in different way to change it now.
> 
> As @raon0211 [commented](https://github.com/toss/slash/pull/196#issuecomment-1402040307), I think currently we should not make a breaking change without necessary, and apparent needs or use cases, like your changes.
> 
> So I think we will close this PR. Any idea?

Okay, when useErrorBoundary need to have new feature of return type or if @toss/error-boundary need to make any other BREAKING CHANGE, I'll come back with PR👍

manudeli:
> @manudeli I agree with [your comment](https://github.com/toss/slash/pull/196#issuecomment-1402163796) about maintaining consistency and confusing of naming, but I think it is currently inappropriate to merge this change.
> 
> Because this is a breaking change. I think the scalability and the consistency that you commented should be considered at a draft level. I think currently `useErrorBoundary` is at a maintaining level. So I'm sorry but we need to approach in different way to change it now.
> 
> As @raon0211 [commented](https://github.com/toss/slash/pull/196#issuecomment-1402040307), I think currently we should not make a breaking change without necessary, and apparent needs or use cases, like your changes.
> 
> So I think we will close this PR. Any idea?

Okay, when useErrorBoundary need to have new feature of return type or if @toss/error-boundary need to make any other BREAKING CHANGE, I'll come back with PR👍

PR #200, opened 2023-01-25
perf(error-boundary): optimize return of useErrorBoundaryGroup

manudeli:
## Overview
Optimize return of useErrorBoundaryGroup by using useMemo

### AS-IS
```tsx
const Comp = () => {
  // useErrorBoundaryGroup make new object, whenever Comp is rendered, for return type(group)
  const group = useErrorBoundaryGroup 

  ...
}
```

### TO-BE
```tsx
const Comp = () => {
  // useErrorBoundaryGroup don't make new object, whenever Comp is rendered, for return type(group) 
  const group = useErrorBoundaryGroup 

  ...
}
```

## PR Checklist

### DIFF
--- a/packages/react/error-boundary/src/ErrorBoundaryGroup.spec.tsx
+++ b/packages/react/error-boundary/src/ErrorBoundaryGroup.spec.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { render, renderHook, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import ErrorBoundary from './ErrorBoundary';
 import { ErrorBoundaryGroup, useErrorBoundaryGroup } from './ErrorBoundaryGroup';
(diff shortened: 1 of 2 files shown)


ChanhyukPark-Tech:
Thank you for your good opinion!
I have a question.

You are currently returning an object containing only reset.
If you look at the optimization you did, each time the reset changes, a new object is created. In an object that only has reset, an object is created every time the reset is changed, so is it necessary to optimize it?

Rather, I think it will cost more to optimize!

If more expensive logic is included in the object in the future, it will be effective, but at the moment, it will have a negative effect, what do you think?!

>Memoization: an optimization technique used primarily to speed up computer programs by storing the results of expensive function calls and returning the cached result when the same inputs occur again.

manudeli:
> Thank you for your good opinion! I have a question.
> 
> You are currently returning an object containing only reset. If you look at the optimization you did, each time the reset changes, a new object is created. In an object that only has reset, an object is created every time the reset is changed, so is it necessary to optimize it?
> 
> Rather, I think it will cost more to optimize!
> 
> If more expensive logic is included in the object in the future, it will be effective, but at the moment, it will have a negative effect, what do you think?!
> 
> > Memoization: an optimization technique used primarily to speed up computer programs by storing the results of expensive function calls and returning the cached result when the same inputs occur again.

I intended that useErrorBoundaryGroup returning object's reference need to be same until It need to be refresh during rendering.
So I add test case to prove it. thanks for your helping me to add test code(@ChanhyukPark-Tech)

ChanhyukPark-Tech:
It's a good idea to keep the same reference it is semantically effective! It was nice to have a useful meeting together!

raon0211 (approved):
Thanks!

PR #203, opened 2023-01-28
refactor(react): change management useBooleanState

tooooo1:
## Overview

I changed the method of managing the state of useboolean. I adjusted the boolean value in three ways: setTrue, setFalse, and toggle, but I succeeded in unifying it with one setBool. Users no longer have to call in three things and can control the state with just one.

**The existing one is as below**
```tsx
const [bool, setTrue, setFalse, toggle] = useBooleanState();

setFalse() // change false
setTrue() // change true
toggle() // toggle

bool // use

```

**The improvement is as follows**
```tsx
const [bool, setBool] = useBooleanState();

setBool(false) // change false
setBool(true) // change true
setBool() // toggle

bool // use

```

I revised the test code accordingly

image

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useBooleanState.ts
+++ b/packages/react/react/src/hooks/useBooleanState.ts
@@ -14,7 +14,10 @@ import { useCallback, useState } from 'react';
 export const useBooleanState = (defaultValue = false): readonly [boolean, (state?: boolean) => void] => {
   const [bool, setBool] = useState(defaultValue);
 
-  const boolValue = useCallback((state?: boolean) => setBool(b => (typeof state === 'boolean' ? state : !b)), []);
+  const handleBoolChange = useCallback(
+    (state?: boolean) => setBool(b => (typeof state === 'boolean' ? state : !b)),
+    []
+  );
 
-  return [bool, boolValue] as const;
+  return [bool, handleBoolChange] as const;
 };
--- a/packages/react/react/src/hooks/useBooleanState.ts
+++ b/packages/react/react/src/hooks/useBooleanState.ts
@@ -5,26 +5,16 @@ import { useCallback, useState } from 'react';
  * boolean 타입으로 useState를 쉽게 사용할 수 있는 hook 입니다.
  *
  * ```ts
- * function useBooleanState(defaultValue = false): readonly [bool, setTrue, setFalse, toggle];
+ * function useBooleanState(defaultValue = false): readonly [bool, boolValue];
  * ```
  *
  * @example
- * const [open, openBottomSheet, closeBottomSheet, toggleBottomSheet] = useBooleanState(false);
+ * const [open, setOpen] = useBooleanState(false);
  */
-export const useBooleanState = (defaultValue = false): readonly [boolean, () => void, () => void, () => void] => {
+export const useBooleanState = (defaultValue = false): readonly [boolean, (state?: boolean) => void] => {
   const [bool, setBool] = useState(defaultValue);
 
-  const setTrue = useCallback(() => {
-    setBool(true);
-  }, []);
+  const boolValue = useCallback((state?: boolean) => setBool(b => (typeof state === 'boolean' ? state : !b)), []);
 
-  const setFalse = useCallback(() => {
-    setBool(false);
-  }, []);
-
-  const toggle = useCallback(() => {
-    setBool(b => !b);
-  }, []);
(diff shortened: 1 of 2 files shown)


packages/react/react/src/hooks/useBooleanState.ts
```diff
@@ -5,26 +5,16 @@ import { useCallback, useState } from 'react';
...
+ * const [open, setOpen] = useBooleanState(false);
  */
-export const useBooleanState = (defaultValue = false): readonly [boolean, () => void, () => void, () => void] => {
+export const useBooleanState = (defaultValue = false): readonly [boolean, (state?: boolean) => void] => {
   const [bool, setBool] = useState(defaultValue);
 
-  const setTrue = useCallback(() => {
-    setBool(true);
-  }, []);
+  const boolValue = useCallback((state?: boolean) => setBool(b => (typeof state === 'boolean' ? state : !b)), []);
```
ChanhyukPark-Tech, line 17: How about changing the function called `boolValue` to like `setBool`?
Isn't the purpose of the function to change the bool value?

tooooo1: @ChanhyukPark-Tech Thank you for telling me about missed.

manudeli:
Although this is BREAKING CHANGE, I could understand why you add this Pull Request, if any user use want to only setFalse or toggle, not setTrue, then they can not avoid to make useless setTrue like below code.

```tsx
const [isBoolean, _, setFalse, toggle] = useBoolean() // useless _ (setTrue)
```

So I want to suggest new useToggle like below, if everyone approve this suggestion with BREAKING CHANGE

```tsx
function Example() {
  const [isFlag, flag] = useBoolean()

  return (
    <>
      <button onClick={flag.toTrue}>turn on</button>
      <button onClick={flag.toFalse}>turn off</button>
      <button onClick={flag.toggle}>toggle</button>
      {isFlag ? 'on' : 'off'}
    </>
  )
}
```

###

raon0211:
In this case, what do you think is the benefit to use `useBooleanState` instead of `useState`? 
The purpose of `useBooleanState` was to make codes like this simple.

```tsx
function SomePage() {
  const [isModalOpen, setIsModalOpen] = useState(false);

  const openModal = () => {
    setIsModalOpen(true);
  };

  const closeModal = () => {
    setIsModalOpen(false);
  }

  return (
    <>
      {/* some modal ... */}
      <button onClick={() => openModal()}>
        Open modal
      </button>
      <Modal isOpen={isModalOpen} onClose={closeModal} />
    </>
  );
}
```

As you see, turning boolean flags on/off is very common when developing React applications, so we have added `useBooleanState` as a utility hook.

tooooo1:
@raon0211 
Thanks for letting us know through examples.
I agree! Here's the flow I was thinking of before blowing up the PR.

As you said, I thought of the advantages of useBooleanState compared to useState, and those advantages naturally led to toggle. Because I didn't think of modal when I saw booleanState. To use the toggle I had to use something like this:

`const [bool, _, _, toggle] = useBooleanState();`

And I looked for a separate utility hook for the toggle, but it wasn't, so I figured I'd have to use the toggle in SLASH via that hook. In order to use the toggle, I had to connect to the 4th depth as above, so I made a PR code like this to improve this.

raon0211:
Yes, I agree that skipping two functions returned by `useBooleanState()` can be annoying. But I'm afraid in this pull request, we're merely returning `value` and `setState` returned from `useState`. I guess this adds no value to why we use `useBooleanState`.

raon0211:
Yes, I understand that calling `toggle` in `useBooleanState` might be awkward. 

However `toggle` is not the main use case of `useBooleanState`; instead making it easy to `setTrue`/`setFalse` is our main use case. I'm afraid this pull request will make main use cases (`setTrue` / `setFalse`) harder. In addition, it will break our API stability.

Why not change the API of `useBooleanState`, but add a new hook like `useToggleState`?

tooooo1:
wow! As I kept thinking about it, I understood what the purpose of this hook was. I'm only getting too hung up on using toggles through arrays. As you said, I think rename would be easier through that hook. thank you for telling me.

HoseungJang:
@tooooo1 I think this PR can be closed. Please open new PR if you want to add a new hook `useToggleState`. Thanks to your great contribution.

PR #204, opened 2023-01-29
refactor(react): use force update return type

tooooo1:
## Overview

removed "as" for type returned by function.
 - didn't have to use type assertion, so removed it.
 - The note sentence has been modified more clearly.

## PR Checklist


packages/react/react/src/hooks/useForceUpdate.ts
```diff
@@ -7,10 +7,10 @@ const updater = (num: number): number => (num + 1) % 1_000_000;
  * @description
  * 반환된 함수를 실행 시 강제로 리렌더가 실행됩니다.
  *
- * @note useReduce가 state보다 조금 더 저렴(?) 하다고 합니다. https://github.com/streamich/react-use/pull/837
+ * @note useReduce가 state보다 가볍다는 의견이 있습니다. https://github.com/streamich/react-use/pull/837
```
manudeli, line 10: How about using `@link`
[jsdoc - @link](https://jsdoc.app/tags-inline-link.html) 

```suggestion
 * @note {@link https://github.com/streamich/react-use/pull/837 useReduce가 state보다 가볍다는 의견이 있습니다.}
```

packages/react/react/src/hooks/useForceUpdate.ts
```diff
@@ -7,10 +7,10 @@ const updater = (num: number): number => (num + 1) % 1_000_000;
...
  */
-export function useForceUpdate() {
+export function useForceUpdate(): () => void {
   const [, forceUpdate] = useReducer(updater, 0);
 
-  return forceUpdate as () => void;
+  return forceUpdate;
```
manudeli, line 15: 1. I guessed maintainer want to use meaning([Type Assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions)) of `as` to override type React.DispatchWithoutAction (ReturnType<typeof useReducer>[1]) for user

2. and also if use `as`, overriding type can be occured relative code than `: () => void`

packages/react/react/src/hooks/useForceUpdate.ts
tooooo1: Thank you for such a great opinion. reflected it.

packages/react/react/src/hooks/useForceUpdate.ts
tooooo1: The reason for the PR I posted is as follows.

1. Override to () => void without Type Assertions.
2. "Type Assertions" forces type information on the type system and skips the type check. Override's role is possible without a "Type Assertions"

-> I decided that there was no reason to choose such a powerful "Type Assertions"

raon0211 (approved):
Thanks!

PR #205, opened 2023-01-30
docs(README.md): add banner

manudeli:
## Overview
add beautiful animated banner gif file in README.md
I converted doc's mp4 file to gif to add in README.md
and then add it in README.md using raw file in GitHub.

banner

## References
I think bigger and animated picture can make bigger impact for user(developers)
like [react-hook-form](https://github.com/react-hook-form/react-hook-form), [TanStack/query](https://github.com/TanStack/query)

image

image

## PR Checklist


BasixKOR:
I think the added banner looks nice, Is there a particular reason to convert the banner into GIF rather than using the original MP4 as a source?

See also: [Replace animated GIFs with video for faster page loads](https://web.dev/replace-gifs-with-videos/)

manudeli:
## Problem of embeded video of README.md
1. It's not really fast to load than gif in README.md
2. this gif image size is smaller than original video. (I actually compressed video to reduce size of gif)
3. No autoplay. (Is there any controller to autoplay video in github? I couldn't find it.)
4. Default layout of embed video is ugly. (file name, video controller, layout line, etc. will be exposed)
5. Controlling as html tag img:src="~.gif" is easier than current GitHub embeded video.

### Directly and Compare it please.
I made [example repo (See this please)](https://github.com/manudeli/video-readme-test) using video in README by drag&drop to prove why this is better

chrome-capture-2023-1-1

I tried to use video but I couldn't find other beautiful solution to add video in README.md than using gif.
If you have alternative solution, could you suggest for me?

HoseungJang:
> 1. This gif image size is smaller than original video. (I actually compressed video to reduce size of gif)
> 2. No autoplay. (Is there any controller to autoplay video in github? I couldn't find it.)
> 3. Ugly default layout of embed video. (file name, video controller, layout line, etc. will be exposed)
> 4. Easier controlling as html tag img:src="~.gif" than current GitHub embeded video.

- The compression algorithm of GIF is very inefficient than MP4.
- The GIF you added has very lower quality compared to the original MP4.
    - I think it is because you tried to lower the size of the GIF too much.
    - But there is only 0.1MB difference between the GIF and the MP4.
- MP4 can be streamed but GIF cannot be streamed. So user should download a whole file if we use GIF.
- ...

So I think GIF has more problems than MP4.

And I think we should use a static image file to add a banner to the top of README, like [react-hook-form](https://github.com/react-hook-form/react-hook-form), [TanStack/query](https://github.com/TanStack/query).

manudeli:
> > 1. This gif image size is smaller than original video. (I actually compressed video to reduce size of gif)
> > 2. No autoplay. (Is there any controller to autoplay video in github? I couldn't find it.)
> > 3. Ugly default layout of embed video. (file name, video controller, layout line, etc. will be exposed)
> > 4. Easier controlling as html tag img:src="~.gif" than current GitHub embeded video.
> 
> * The compression algorithm of GIF is very inefficient than MP4.
> * The GIF you added has very lower quality compared to the original MP4.
>   
>   * I think it is because you tried to lower the size of the GIF too much.
>   * But there is only 0.1MB difference between the GIF and the MP4.
> * MP4 can be streamed but GIF cannot be streamed. So user should download a whole file if we use GIF.
> * ...
> 
> So I think GIF has more problems than MP4.
> 
> And I think we should use a static image file to add a banner to the top of README, like [react-hook-form](https://github.com/react-hook-form/react-hook-form), [TanStack/query](https://github.com/TanStack/query).

I got it. so I changed banner from gif to png in [7c42ed8](https://github.com/toss/slash/pull/205/commits/7c42ed8adf5ea4c868eb124e6b3122c45aaccc4e)

It's so sad that the animation is gone🥲,
but big image also will be much much better than no banner.👍
I'm so glad to add this banner image too.

banner

manudeli:
> > 1. This gif image size is smaller than original video. (I actually compressed video to reduce size of gif)
> > 2. No autoplay. (Is there any controller to autoplay video in github? I couldn't find it.)
> > 3. Ugly default layout of embed video. (file name, video controller, layout line, etc. will be exposed)
> > 4. Easier controlling as html tag img:src="~.gif" than current GitHub embeded video.
> 
> * The compression algorithm of GIF is very inefficient than MP4.
> * The GIF you added has very lower quality compared to the original MP4.
>   
>   * I think it is because you tried to lower the size of the GIF too much.
>   * But there is only 0.1MB difference between the GIF and the MP4.
> * MP4 can be streamed but GIF cannot be streamed. So user should download a whole file if we use GIF.
> * ...
> 
> So I think GIF has more problems than MP4.
> 
> And I think we should use a static image file to add a banner to the top of README, like [react-hook-form](https://github.com/react-hook-form/react-hook-form), [TanStack/query](https://github.com/TanStack/query).

I got it. so I changed banner from gif to png in [7c42ed8](https://github.com/toss/slash/pull/205/commits/7c42ed8adf5ea4c868eb124e6b3122c45aaccc4e)

It's so sad that the animation is gone🥲,
but big image also will be much much better than no banner.👍
I'm so glad to add this banner image too.

banner

raon0211 (approved):
Thanks, @manudeli !

PR #209, opened 2023-01-31
feat(@toss/storage): Gave scalability to two methods of NumberTypedStorage.

ChanhyukPark-Tech:
## Overview

Building an application, you might want to increase and decrease the number as much as you want, instead of adjusting the number one by one.
In that case, two methods(increase, decrease) were modified to receive argument.

## PR Checklist


packages/common/storage/src/typed/storages/NumberTypedStorage.ts
```diff
@@ -2,13 +2,13 @@
...
     const value = this.get() ?? 0;
-    this.set(value + 1);
+    this.set(value + offset);
   }
 
-  public decrease(): void {
+  public decrease(offset = 1): void {
     const value = this.get() ?? 0;
-    this.set(value - 1);
+    this.set(value - offset);
```
HoseungJang, line 12: I hope this will be added to a new method `add` and `subtract`.

ChanhyukPark-Tech: It might be a better solution to break it down into several functions instead of just implementing in just one function(`increase`) it's a good solution.

However, in the case of the `Increase` function, I think It can be better get offset as a parameter. The reason why I think is as follows.

1. We don't know that the function name increase itself increases by 1(only just 1). who can think increase 10 or 100

2. Since the name `add` is to add two numbers, what and what do you add when you call add just one parameter like `add(10)`? This can lead to confusion.

```tsx

// when make add function
const INCREASE_OFFSET = 5;

const storage = createTypedLocalStorage('key', {
  initialValue: 0,
});

storage.add(INCREASE_OFFSET);

// when make increase function to accept parameter
const INCREASE_OFFSET = 5;

const storage = createTypedLocalStorage('key', {
  initialValue: 0,
});

storage.increase(INCREASE_OFFSET);

```

Looking at the code, I think both are okay, so I'd like to hear from other people.

HoseungJang: Yes I have thought about it again, your changes look great.

packages/common/storage/src/typed/storages/NumberTypedStorage.ts
```diff
@@ -2,13 +2,13 @@
 import { TypedStorage } from './TypedStorage';
 
 export class NumberTypedStorage extends TypedStorage<number> {
-  public increase(): void {
+  public increase(offset = 1): void {
```
HoseungJang, line 5: Please specify the type of `offset`.
```typescript
public increase(offset: number = 1): void;
```

packages/common/storage/src/typed/storages/NumberTypedStorage.spec.ts
```diff
@@ -1,19 +1,33 @@
 import { NumberTypedStorage } from './NumberTypedStorage';
 
-describe('NumberTypedStorage는', () => {
-  describe('increase() 메소드로', () => {
-    it('주어진 key의 값을 1 증가시킨다.', () => {
-      const typed = new NumberTypedStorage('increase-test', { initialValue: 10 });
+describe('NumberTypedStorage', () => {
+  describe('increase() method', () => {
+    it('should increase the value of the given key by 1. when given no argument', () => {
```
HoseungJang, line 5:
```typescript
describe('when there is no given offset', () => {
  it('should increase the value of the given key by 1', () => {
    ...
  });
});
```

packages/common/storage/src/typed/storages/NumberTypedStorage.spec.ts
```diff
@@ -1,19 +1,33 @@
...
+      const typed = new NumberTypedStorage('increase-test-no-argument', { initialValue: 10 });
       typed.increase();
       expect(typed.get()).toBe(11);
     });
+    it('should increase the value of the given key by argument.', () => {
```
HoseungJang, line 10:
```typescript
describe('when there is given offset', () => {
  it('should increase the value of the given key by offset', () => {
    ...
  });
});
```

packages/common/storage/src/typed/storages/NumberTypedStorage.spec.ts
```diff
@@ -1,19 +1,33 @@
...
+      typed.increase(10);
+      expect(typed.get()).toBe(30);
+    });
   });
 
-  describe('decrease() 메소드로', () => {
-    it('주어진 key의 값을 1 감소시킨다.', () => {
-      const typed = new NumberTypedStorage('decrease-test', { initialValue: 10 });
+  describe('decrease() method', () => {
+    it('should decrease the value of the given key by 1. when given no argument', () => {
```
HoseungJang, line 20:
```typescript
describe('when there is no given offset', () => {
  it('should decrease the value of the given key by 1', () => {
    ...
  });
});
```

packages/common/storage/src/typed/storages/NumberTypedStorage.spec.ts
```diff
@@ -1,19 +1,33 @@
...
+      const typed = new NumberTypedStorage('decrease-test-no-argument', { initialValue: 10 });
       typed.decrease();
       expect(typed.get()).toBe(9);
     });
+    it('should decrease the value of the given key by argument.', () => {
```
HoseungJang, line 25:
```typescript
describe('when there is given offset', () => {
  it('should decrease the value of the given key by offset', () => {
    ...
  });
});
```

packages/common/storage/src/typed/storages/NumberTypedStorage.ts
ChanhyukPark-Tech: It's good that it's explicit! but it is ok to ignore line warning?

image

ChanhyukPark-Tech:
@HoseungJang thank you for your sophisticated translation😁

HoseungJang (approved):

PR #210, opened 2023-01-31
fix(@toss/storage): remove duplicated test code

ChanhyukPark-Tech:
## Overview

There are two test codes that verify the same logic, so I removed one.

## PR Checklist


HoseungJang (approved):
Thanks.

PR #211, opened 2023-01-31
fix(react): useCombinedRefs

tooooo1:
# fix(react): useCombinedRefs

### Problem 1

image

### problem 1 Solved

Added import React.

### Problem 2

image

I ran the test as per example. And it failed and discovered there was a problem.

### Problem 2 Solved

image

test code -> Added undefined

### Problem 3

image

The defined `CallbackRef` type does not match. And inside `Ref`, there is already a Refcallback type.

### Problem 3 Solved

image

remove CallbackRef

### Problem 4

image

The return type was too wide as Ref.

### Problem 4 Solved

image

I solved it by narrowing down the type to Refcallback inside the Ref type.

### all problems solved

It solved all the problems in turn. also added test cases

( + Finally I changed `for of` to `forEach`...!)

## PR Checklist


raon0211:
Why do you think readability is improved here?

tooooo1:
I thought I could focus on the logic below through a early return...!

HoseungJang:
> I thought I could focus on the logic below through a early return...!

What is the logic you think? I think it is to update all refs depending on whether it is an object or a callback, and it is clear without this change.

tooooo1:
@raon0211  @HoseungJang 
Review again please

packages/react/react/src/hooks/useCombinedRefs.ts
```diff
@@ -18,17 +16,15 @@ type CallbackRef<T> = (ref: T | null) => void;
...
+      refs.forEach(ref => {
         if (typeof ref === 'function') {
           ref(value);
         } else if (ref != null) {
           (ref as MutableRefObject<T>).current = value;
         }
-      }
-    },
+      }),
     [refs]
```
manudeli, line 29: @tooooo1 
refs will be always new array. so I spread it to optimize correctly.
and then I added eslint-disable-next-line.

chakra-ui do this to optimize correctly too. you can check it in [@chakra-ui's useMergeRefs](https://github.com/chakra-ui/chakra-ui/blob/main/packages/hooks/use-merge-refs/src/index.ts)

```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
    refs
```

manudeli (changes requested):
I added some code to optimize useCombinedRefs.

packages/react/react/src/hooks/useCombinedRefs.spec.tsx
```diff
@@ -1,28 +1,30 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { useCallback, useRef } from 'react';
+import React, { useCallback, useRef } from 'react';
```
HoseungJang, line 3: I think you don't have to import `React` because `tsconfig.build.json` ignores `*.spec.tsx`. There is no problem to run test.

packages/react/react/src/hooks/useCombinedRefs.ts
HoseungJang: Strictly, I think this is not optimization. I think optimization is improving performance of something.

With or without your suggestion, there's not any performance improvement of `useCombinedRefs`.

But, with your suggestion, the potential that the reference of the returned callback of `useCombinedRefs` changed on every render will be disappeared and developers won't have to memoize the `refs` parameter.

packages/react/react/src/hooks/useCombinedRefs.ts
```diff
@@ -18,17 +16,16 @@ type CallbackRef<T> = (ref: T | null) => void;
  *   return <div ref={ref} />;
  * })
  */
-export function useCombinedRefs<T>(...refs: Array<Ref<T> | CallbackRef<T>>): Ref<T> {
-  return useCallback(
-    (value: T) => {
-      for (const ref of refs) {
+export const useCombinedRefs = <T>(...refs: Array<Ref<T>>): RefCallback<T> =>
```
HoseungJang, line 19: Why did you change function declaration into arrow function?

packages/react/react/src/hooks/useCombinedRefs.spec.tsx
raon0211: I think fixing it like [this](https://github.com/toss/slash/pull/212) would be a better option.

packages/react/react/src/hooks/useCombinedRefs.ts
raon0211: I think specifying the return type to be `RefCallback<T>`, rather than `Ref<T>`, exposes the actual implementation, making us hard to change the internal implementation. Were there some backgrounds to make return types more specific?

packages/react/react/src/hooks/useCombinedRefs.ts
```diff
@@ -18,17 +16,16 @@ type CallbackRef<T> = (ref: T | null) => void;
...
+      refs.forEach(ref => {
         if (typeof ref === 'function') {
           ref(value);
         } else if (ref != null) {
           (ref as MutableRefObject<T>).current = value;
         }
-      }
-    },
-    [refs]
+      }),
```
raon0211, line 28: I think using for-loops or Array#forEach is a matter of choice and style, and prefer to keep the original code unless there are some specific reasons. 

If they change a lot, there would be many minor changes in the pull request, making it hard to see the purpose of the pull request (because it contains a lot of unrelated changes). Also, in the long run, it will make it hard to track the changes and history of the code. What do you think?

packages/react/react/src/hooks/useCombinedRefs.spec.tsx
HoseungJang: Yes I think so. `tsconfog.build.json` should be only used to build.

I was going to fix it but you already did.

packages/react/react/src/hooks/useCombinedRefs.ts
HoseungJang: I agree with @raon0211 .

In the case of `useCombinedRefs`, I think there is no dramatic performance improvement between forEach and for loop. So if there is no specific reason to change it, we should keep for-loop.

HoseungJang: In addition, I don't know what the actual problem this PR is solving.

tooooo1: First of all, I tried to run the hook in the test code as in the example, but an error occurred.

Second, CallbackRef is newly defined, which is already defined as RefCallback in the Ref type.

Am I thinking wrong?

tooooo1: And although you already solved it for me, I was able to find the same problem as [this](https://github.com/toss/slash/pull/212).

HoseungJang: This is not the type error because of `CallbackRef` or because we didn't allow undefined.

`useCombinedRefs` is a hook to update the passed refs with one same value.

But, in the screenshot you attached, the passed refs to `useCombinedRefs` are not accepting the same type of the value. Therefore, the type error should be occurred.

I'm sorry but I think your case is incorrect and the hook works well.

HoseungJang: To sum up, all passed refs should accept the same type. So they can be updated at the same time by `useCombinedRefs`.

tooooo1: Oh, I see, I think I'm misunderstanding. Thanks for telling me. Another story, but I have an additional question.

I was using type assertion but I took refObject readonly and changed it to MutableRefObject. I had only bad thoughts about type assertions, but this code changed my mind. am i right?

If so, it's so cool and I wonder why you're doing it this way. Because we know that relying on type inference and type checking will make your code safer and more maintainable in the long run. We also know that it helps us detect potential issues early in the development process. So I wonder why they made this change!

packages/react/react/src/hooks/useCombinedRefs.spec.tsx
```diff
@@ -5,24 +5,26 @@ import { useCombinedRefs } from './useCombinedRefs';
...
+      const ref1 = useRef<HTMLDivElement>();
+      const ref2 = useRef();
+      const ref3 = useRef(null);
+      const ref4 = useRef<HTMLDivElement>(null);
+      const ref5 = useCallback((element: HTMLDivElement | null) => {
+        refCallback(element);
       }, []);
 
-      const ref = useCombinedRefs(ref1, ref2, ref3);
+      const ref = useCombinedRefs(ref1, ref2, ref3, ref4, ref5);
```
HoseungJang, line 20: https://github.com/toss/slash/pull/211#discussion_r1097054749

So please revert this change.

packages/react/react/src/hooks/useCombinedRefs.ts
tooooo1: Also, I used the example stated in the jsdoc, not the example I used. If so, is it correct to change the jsdoc correctly?

tooooo1: thank you @HoseungJang 

I'm here because I want to contribute, but I'm just learning and going. So what should I do with this PR?

raon0211: I think why not close this pull request, and go on to the next task? 🚀

tooooo1: good @raon0211

tooooo1: My challenge for SLASH continues!!

PR #214, opened 2023-02-08
docs(react): fix example of useCombinedRefs's js-doc

manudeli:
## Overview
@HoseungJang, I found you want to prevent that useCombinedRefs accept MutableRefObject in https://github.com/toss/slash/pull/211#discussion_r1097071688. so I understood why useCombinedRefs not accepting MutableRefObject is to prevent ref.current's mutation by readonly field of RefObject, making TypeScript error if trying mutation of ref.current.

But, in example, useCombinedRefs accept MutableRefObject, so I fixed it.

also Thanks to @tooooo1 that make this https://github.com/toss/slash/pull/211

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useCombinedRefs.ts
+++ b/packages/react/react/src/hooks/useCombinedRefs.ts
@@ -11,7 +11,7 @@ type CallbackRef<T> = (ref: T | null) => void;
  *
  * @example
  * const SomeComponent = forwardRef((props, parentRef) => {
- *   const myRef = useRef();
+ *   const myRef = useRef(null);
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
👍

HoseungJang:
> @HoseungJang, I found you want to prevent that useCombinedRefs accept MutableRefObject in https://github.com/toss/slash/pull/211#discussion_r1097071688.

No, I mean, **in React type definition**, only `MutableRefObject` can be changed and `RefObject` can't be changed.

So I worried about if it is the only restriction in type definition, or if runtime errors could be occurred.

But I think the former is right.

PR #215, opened 2023-02-21
fix(utils): remove the type that deleted the symbol which name is `ObjectKeys`.

ChanhyukPark-Tech:
## Overview

Although the object's key value could be a symbol, the array of keys was blocked from entering the symbol.

I think I've limited it in the past because of [this issue](https://github.com/microsoft/TypeScript/issues/24622), but now the keyof command works well for the object to which the symbol belongs.

If there's any other reason why `pick` function blocked the symbol, please tell me!

## Screenshot in my vscode

image

## PR Checklist

### DIFF
--- a/packages/common/utils/src/object/pick.spec.ts
+++ b/packages/common/utils/src/object/pick.spec.ts
@@ -19,4 +19,22 @@ describe('pick', () => {
       JP: 'JP',
     });
   });
+  it('should create an object composed of the given symbol keys', () => {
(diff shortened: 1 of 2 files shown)


HoseungJang (approved):
Thanks!

I think we need `ObjectKeys` type when we use some utils which exclude Symbol like `Object.keys`.

But in `pick`, you are right. We don't have to exclude Symbol.

PR #218, opened 2023-02-23
docs(@toss/uitls): add NonEmptyArray example

hhhminme:
## Overview
image
image

- 영어문서와 한글 문서의 내용이 달라 통일해주었습니다.
- 타입 형태와 example이 빠져 있어 예시 코드를 추가해주었습니다. 
- `isNonEmptyArray` util에서 해당 타입을 사용하고 있어 `isNonEmptyArray`와 관련된 설명과 링크를 추가해주었습니다.

## PR Checklist

### DIFF
--- a/packages/common/utils/src/array/NonEmptyArray.en.md
+++ b/packages/common/utils/src/array/NonEmptyArray.en.md
@@ -1,3 +1,25 @@
 # NonEmptyArray
 
-A type which indicates that an array is not empty. We can assert an array not to be empty by `isNonEmptyArray` function.
+A type which indicates that an array is not empty.
+
+```typescript
+type NonEmptyArray<T> = [T, ...T[]];
+```
+
+- You can use this to create a function like [isNonEmptyArray](https://slash.page/ko/libraries/common/utils/src/array/isnonemptyarray.i18n/) that checks if an array has at least one element.
+
+# Example
+
+```typescript
+// There is no problem at the time of declaration, but runtime errors may occur depending on the internal logic.
+const getSum = (array: number[]) => array.reduce((x, y) => x + y);
+
+// [ERR]: Reduce of empty array with no initial value
+getSum([]);
+
+// Runtime errors can be prevented through type errors at the time of declaration.
+const getSum = (array: NonEmptyArray<number>) => array.reduce((x, y) => x + y);
+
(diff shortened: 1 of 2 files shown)


ChanhyukPark-Tech (approved):
cool..
docs written well😁

hhhminme:
Thank you for the quick review!

But I have a question. Is there a special reason for communicating in English in this library?👀

ChanhyukPark-Tech:
>Is there a special reason for communicating in English in this library?👀

In [contribution docs](https://github.com/toss/slash/blob/main/.github/CONTRIBUTING.md)
written `All communications in this repo will be by English.`

So I think you should revise your PR content, too!

When I talked to the library owner in the past,
Since slash is an open source, non-Korean users often use it, and he said that they thought it would be convenient if it was written in English, a common language, when grasping the history of slash.

hhhminme:
Thank you for your kind explanation!
I'll fix it!

packages/common/utils/src/array/NonEmptyArray.en.md
```diff
@@ -1,3 +1,25 @@
...
+
+// [ERR]: Reduce of empty array with no initial value
+getSum([]);
+
+// Runtime errors can be prevented through type errors at the time of declaration.
+const getSum = (array: NonEmptyArray<number>) => array.reduce((x, y) => x + y);
+
+// Argument of type '[]' is not assignable to parameter of type 'NonEmptyArray<number>'. Source has 0 element(s) but target requires
+getSum([]);
+```
```
raon0211, line 25: Since `sum` in our libraries are defined as `arr.reduce((x, y) => x + y, 0)`, I thought this example might be a little confusing. 

Why not use a more easy example like `getFirstElement()`? 

```ts
function getFirstElement<T>(arr: T[]): T {
  return arr[0];
}

const items: number[] = [];

// Runtime error
getFirstElement([]).toString()
```

hhhminme: @raon0211 thank you I thought it was too difficult!
I modified the code as below and added it to the example.
```typescript
function getFirstElement<T>(arr: T[]): T {
  return arr[0];
}

const items: number[] = [];

// Runtime error
getFirstElement([]).toString();

// Type error
const fixItems: NonEmptyArray<number> = [];
```
Thank you for checking.

raon0211 (approved):

PR #219, opened 2023-02-26
docs(i18n):  Edit banner style

hhhminme:
## Overview

image

image

- The part where the banner looks different on the English page and the Korean page is unified to match the English page.
- The reason why I changed it to English page is because the commit log of the banner addition code is more recent. (I think the English page is prettier aesthetically 👀)
## PR Checklist


raon0211 (approved):
Thanks, @hhhminme !

Issue #198, opened 2023-01-24
[Feature]: FullHeight container's element could be other than just div


ChanhyukPark-Tech:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: @toss/emotion-utils

### **Overview**

```tsx
// FullHeight.tsx

export function FullHeight({ children, ...props }: { children: ReactNode } & HTMLAttributes<HTMLDivElement>) {

  return (
    <div> // why only div tag?
      {children}
    </div>
  );
}


```

If you look at the current implementation of FullHeight, it is fixed and wrapped with a `div` container.

However, in actual development, it is often necessary to semantically use `section` or `main` and other elements instead of div containers. In that case, it is inconvenient to use FullHeight. 

I think it would be good to have better scalability!

### **Describe the solution you'd like**

like [`OpenGraph`](https://slash.page/libraries/react/react/src/components/opengraph/opengraph.tsx.tossdocs/)

```tsx
interface Props {
  container?: ComponentType<any>; // accept other tag
}
```

manudeli:
How about `as` prop for that feature(aka. polymorphic component)?

### Proper prop name to provide polymorphic feature
I think Opengraph's container is proper for its meaning because default container is Fragment,
so if anyone use container prop, container will contain children(for opengraph tags).

but I think FullHeight is different case. Default element need to be div like current implementation.
So if you want to add polymorphic feature, I want to suggest that prop name as `as` like Box's as prop of @chakra-ui/react

### More feature expected
If possible, I really expect situation using `@toss/emotion-utils` using `@toss/framer-motion`'s motion.div as 'as prop'.
like Box of `@chakra-ui/react` using motion.div as `as` prop

#### Example: I imaged smooth shrinking animation of FullHeight for emphasize overlay and so easy implementation.

```tsx
const Page = () => {
  const [isShrinkForOverlay, setIsShrinkForOverlay] = useState(false)
  const overlay = useOverlay() // use @toss/use-overlay

  const openFooConfirmDialog = () =>  new Promise<boolean>(resolve => {
	    overlay.open(({ isOpen, close }) => (
	      <FooConfirmDialog
	        open={isOpen}
	        onClose={() => {
	          resolve(false);
	          close();
	          setIsShrinkForOverlay()
	        }}
	      />
	    ));
	  })

  return (
    <FullHeight as={motion.div} initial={{.scale: 1 }} animate={{.scale: isShrinkForOverlay ? 0.9 : 1 }}>
[...]

PR #221, opened 2023-03-03
fix(emotion-utils): Allows 'as' prop in Flex and Stack to receive all elements

manudeli:
# Flex, Stack's as prop improvement TypeScriptly
before this Pull Request, I thought `Flex, Stack's as prop` is TypeScriptly too free.
I want to fix this TypeScriptly free `Flex, Stack's as prop`

Pull Request about https://github.com/toss/slash/issues/198#issuecomment-1423399747

## After this Pull Request

1. `as prop` will only accept StringElementType(ElementType & string) almost same with keyof JSX.InstrintElements
image

2. `as prop` will restrict to Comoponent can accept proper type ComponentProps 
image

3. `as prop` will restrict to Comoponent can accept proper ref
image

## PR Checklist

6. I have written documents and tests, if needed.


packages/react/emotion-utils/src/Stack.tsx
```diff
@@ -1,50 +1,52 @@
...
+const BaseStack: BaseStackType = forwardRef(function BaseStack<T extends StringElementType = StringElementType>(
+  props: StackProps<T>,
+  ref: ComponentPropsWithRef<T>['ref']
+) {
   const { direction = 'vertical', gutter: gutterSpace = 24, as = 'div', selector, ...rest } = props;
-
   return (
     <Flex
       as={as}
-      ref={ref as any}
```
manudeli, line 27: we can remove `as any` after this Pull Request

packages/react/emotion-utils/src/Stack.tsx
```diff
@@ -1,50 +1,52 @@
...
-interface StackProps<T extends keyof JSX.IntrinsicElements = 'div'>
-  extends AsProps<T>,
-    Omit<FlexOptions, 'direction'>,
-    Omit<GutterOptions, 'direction' | 'space'> {
-  direction?: AxisDirection;
-  gutter?: number;
-}
+type StackProps<T extends StringElementType = StringElementType> = AsProps<T> & StackOptions;
 
-type StackReturnType = <T extends keyof JSX.IntrinsicElements = 'div'>(
```
manudeli, line 17: I renamed `StackReturnType` to `BaseStackType` like `BaseFlexType` because this is not Return type

packages/react/emotion-utils/src/Stack.tsx
```diff
@@ -1,50 +1,52 @@
...
     <Flex
       as={as}
-      ref={ref as any}
+      ref={ref}
       css={gutter(direction, gutterSpace, selector)}
       direction={direction === 'vertical' ? 'column' : 'row'}
       {...rest}
     />
   );
-}) as StackReturnType;
```
manudeli, line 33: I removed unnecessary type assertion

packages/react/emotion-utils/src/types.ts
```diff
@@ -10,6 +10,9 @@ export type AxisDirection = 'vertical' | 'horizontal';
 export type InferenceHTMLElement<K extends keyof JSX.IntrinsicElements> = NonNullable<
   Extract<JSX.IntrinsicElements[K]['ref'], React.RefObject<any>>['current']
 >;
-export interface AsProps<T extends keyof JSX.IntrinsicElements> extends AllHTMLAttributes<T> {
+
+export type StringElementType = ElementType & string;
```
manudeli, line 14: StringElementType can restrict to ElementType can't accept ComponentType, only accept JSX.InstrincElement

```ts
type StringElementType = ElementType & string
```

```ts
type ElementType<P = any> =
  {
    [K in keyof JSX.IntrinsicElements]: P extends JSX.IntrinsicElements[K] ? K : never
  }[keyof JSX.IntrinsicElements] |
  ComponentType<P>;
```

packages/react/emotion-utils/src/types.ts
```diff
@@ -10,6 +10,9 @@ export type AxisDirection = 'vertical' | 'horizontal';
...
+
+export type AsProps<T extends StringElementType> = ComponentPropsWithoutRef<T> & {
```
manudeli, line 16: previously, AllHTMLAttributes make component can accept all
but after this Pull Request AsProp will restrict to component accept only as's componentProps

manudeli (commented):
I added comments to explain why and how I make this change.

packages/react/emotion-utils/src/types.ts
manudeli: I thought StringElementType can be awkward name. so If you(reviewers) think so too, could you guide it to rename?

packages/react/emotion-utils/src/Stack.tsx
```diff
@@ -1,50 +1,52 @@
...
+});
 
-Stack.Vertical = forwardRef<HTMLElement, StackWithDirectionProps>(function StackVertical(props, ref) {
+Stack.Vertical = forwardRef(function StackVertical<T extends StringElementType = StringElementType>(
+  props: StackProps<T>,
+  ref: ComponentPropsWithRef<T>['ref']
+) {
   return <Stack direction="vertical" {...props} ref={ref} />;
-}) as StackReturnType;
+});
```
ChanhyukPark-Tech, line 52: It's not related to this Pull Request, so can we prevent human errors that following example?

direction problem he wants horizontal? or vertical?

```
<Stack.Vertical direction="horizontal">I want horizontal? or vertical?</Stack.Vertical>;
```

Maybe it's in the same context as this https://github.com/toss/slash/pull/149!

manudeli: I wanted to prevent it too, but in this library this maintainer's comment (https://github.com/toss/slash/pull/149#issuecomment-1321333759) will be proper answer for your question.

ChanhyukPark-Tech: I fully understood now thx😃

packages/react/emotion-utils/src/flex.tsx
```diff
@@ -41,19 +41,20 @@ export function flex(
...
-  const Component = as as any;
-
-  return <Component ref={ref} css={flex({ align, direction, justify })} {...rest} />;
-}) as BaseFlexType;
+export const BaseFlex: BaseFlexType = forwardRef(function BaseFlex<T extends StringElementType = StringElementType>(
+  props: FlexProps<T>,
+  ref: ComponentPropsWithRef<T>['ref']
+) {
+  const { as = 'div', direction = 'row', justify = 'flex-start', align = 'stretch', ...rest } = props;
+  const As = as;
```
HoseungJang, line 55: I think we can keep the previous name `Component`. What do the name `As` means here?

packages/react/emotion-utils/src/flex.tsx
```diff
@@ -41,19 +41,20 @@ export function flex(
 
 flex.center = (direction?: FlexOptions['direction']) => flex({ justify: 'center', align: 'center', direction });
 
-interface FlexProps<T extends keyof JSX.IntrinsicElements = 'div'> extends AsProps<T>, FlexOptions {}
+export type FlexProps<T extends StringElementType = StringElementType> = AsProps<T> & FlexOptions;
 
-type BaseFlexType = <T extends keyof JSX.IntrinsicElements = 'div'>(
-  props: FlexProps<T> & { ref?: Ref<InferenceHTMLElement<T>> }
-) => ReactElement | null;
+export type BaseFlexType = <T extends StringElementType = StringElementType>(
```
HoseungJang, line 46: I think this doesn't have to be exported.

HoseungJang, line 46: I think it would be better to name it `FlexComponentType`.

packages/react/emotion-utils/src/flex.tsx
```diff
@@ -63,20 +64,23 @@ type FlexType = typeof BaseFlex & {
...
 ) {
   return <BaseFlex align="center" {...props} ref={ref} />;
 });
 
-Flex.CenterHorizontal = forwardRef<HTMLElement, FlexProps<keyof JSX.IntrinsicElements>>(function CenterHorizontal(
-  props,
-  ref
+Flex.CenterHorizontal = forwardRef(function CenterHorizontal<T extends StringElementType = StringElementType>(
+  props: FlexProps<T>,
+  ref: ComponentPropsWithRef<T>['ref']
```
HoseungJang, line 83: I think the code below would be better.

```typescript
export type FlexProps<T extends StringElementType = StringElementType> = AsProps<T> & FlexOptions;

type FlexComponentType = <T extends StringElementType = StringElementType>(
  props: FlexProps<T> & Partial<Pick<ComponentPropsWithRef<T>, 'ref'>>
) => JSX.Element | null;

function createFlexComponent(flexOptions: FlexOptions): FlexComponentType {
  return forwardRef(<T extends StringElementType = StringElementType>(
    props: FlexProps<T>,
    ref: ComponentPropsWithRef<T>['ref']
  ) => {
    const { as = 'div', direction = 'row', justify = 'flex-start', align = 'stretch', ...rest } = props;
    const As = as;
    return <As ref={ref} css={flex({ ...flexOptions, direction, align, justify })} {...rest} />;
  });
}

type FlexType = FlexComponentType & {
  Center: FlexComponentType;
  CenterVertical: FlexComponentType;
  CenterHorizontal: FlexComponentType;
};

export const Flex = createFlexComponent({}) as FlexType;

Flex.Center = createFlexComponent({ align: 'center', justify: 'center' });

Flex.CenterVertical = createFlexComponent({ align: 'center' });

Flex.CenterHorizontal = createFlexComponent({ justify: 'center' });
```

packages/react/emotion-utils/src/Stack.tsx
HoseungJang, line 52: https://github.com/toss/slash/pull/221#discussion_r1129895450

manudeli (commented):
@HoseungJang Thanks for your advice! I refactored this Pull Request in [454096b](https://github.com/toss/slash/pull/221/commits/454096bc6a9d7f10db79c0e4d4b5802a4969ef12).
It seems to be much better with removing repetition. and You made proper naming Flex, Stack's type.

Thanks for the good advice. However, I'll leave my small editing your advices as comments, so please check if it's okay. Please confirm.

packages/react/emotion-utils/src/flex.tsx
manudeli: Thanks, cool renaming 👍
There is no Base in Flex so FlexComponentType is much better naming I can understand easily too.

packages/react/emotion-utils/src/flex.tsx
manudeli: This is a code I left because I was bored haha😁 The name As is a component of 'as prop', and I was wondering if there was a way to shorten the name. But in the end, I changed it to a commonly used name called Component.

Thanks for your eagle eyes!

HoseungJang (approved):
Sorry I'm late. Thanks!

Issue #220, opened 2023-03-01
[BUG]: useStorageState hook type error


ojj1123:
### **Package Scope**

Package name: @toss/react

#### tl;dr
- In the example of `useStorageState`, a type error occurs. This is because `defaultValue` is inferred as a `literal type` rather than a `primitive type`.
- I tried to infer the type of `defaultValue` without writing a `generic` in `useStorageState`, but I couldn't find a way. Please advise if there is any other way
- If there is no other way, I will add a phrase to the document that a `generic` must be created if the `defaultValue` is a `primitive type` (string, boolean, number).

---

This is an issue about a type error occurring in the `useStorageState` hook.

If you try to use the example in [the document of `useStorageState` hook](https://slash.page/libraries/react/react/src/hooks/usestoragestate/) as it is, the following type error occurs.
```ts
function MyComponent() {
  const [state, setState, refresh] = useStorageState('@service/some-resource', {
    defaultValue: 0,
  });

  useEffect(() => {
    setState(x => x + 1);
    // ^ type error
  }, []);

  ...
}
```
<img width="513" alt="type-error" src="https://user-images.githubusercontent.com/33178048/222148878-8f7e8a6a-18c5-407e-9224-cc22025a3987.png">

However, if you write a `generic`, type errors do not occur:
```ts
const [state, setState, refresh] = useStorageState<number>('@service/some-resource', {
  defaultValue: 0,
});

useEffect(() => {
  // No error
  setState((x) => x + 1);
}, [setState]);
```

I confirmed that the `state` among the return values of `useStorageState` in the example above is not inferred as `number`, but as literal type `0`. In addition, even when the `defaultValue` is a primitive type such as `boolean` or `string`, it is inferred only as a `literal type`:
```ts
  // const state1: true
  const [state1, setState1, refresh1] = useStorageState('boolean', {
    defaultValue: true,
  });

  // const state2: "str"
  const [state2, setState2, refresh2] = useStorageState('string', {
    defaultValue: 'str',
  });
```

**Here, I want to make type inference without writing generic like `React.useState`.** However, because of the type constraint that prevents the `symbol` value from entering the `devalutValue` of the `useStorageState` argument, type inference is not possible as a primitive type like `React.useState`.
```ts
// const count: number
const [count, setCount] = useState(1);

// const state: number  ✅ What I want
// const state: 0  😢 Actually
const [state, setState, refresh] = useStorageState('number', {
  defaultValue: 0
})
```

[...]

PR #223, opened 2023-03-10
fix(react): infer defaultValue as primitive type

Jinho1011:
## defaultValue in useStorageState inferred as a literal type

````typescript
// const state: 0       Before this PR
// const state: number  After this PR
const [state, setState, refresh] = useStorageState('number', {
  defaultValue: 0
})
````

This pull request updates the type definition of defaultValue in codebase to be inferred as a primitive type instead of a literal value. 

To achieve this, I updated the `useStorageState` function's generic type parameter to extends `Serializable` instead of using defaultValue as a `Serializable<T>`.

Pull Request about https://github.com/toss/slash/issues/220

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useStorageState.ts
+++ b/packages/react/react/src/hooks/useStorageState.ts
@@ -19,15 +19,15 @@ interface StorageStateOptionsWithDefaultValue<T> extends StorageStateOptions<T> 
 export function useStorageState<T extends Serializable>(
   key: string
 ): readonly [T | undefined, (value: SetStateAction<T | undefined>) => void, () => void];
-export function useStorageState<T>(
+export function useStorageState<T extends Serializable>(
   key: string,
   { storage, defaultValue }: StorageStateOptionsWithDefaultValue<T>
 ): readonly [T, (value: SetStateAction<T>) => void, () => void];
-export function useStorageState<T>(
+export function useStorageState<T extends Serializable>(
   key: string,
   { storage, defaultValue }: StorageStateOptions<T>
 ): readonly [T | undefined, (value: SetStateAction<T | undefined>) => void, () => void];
-export function useStorageState<T>(
+export function useStorageState<T extends Serializable>(
   key: string,
   { storage = safeLocalStorage, defaultValue }: StorageStateOptions<T> = {}
 ): readonly [T | undefined, (value: SetStateAction<T | undefined>) => void, () => void] {
--- a/packages/react/react/src/hooks/useStorageState.ts
+++ b/packages/react/react/src/hooks/useStorageState.ts
@@ -57,7 +57,7 @@ export function useStorageState<T extends Serializable>(
   const set = useCallback(
     (value: SetStateAction<T | undefined>) => {
       setState(curr => {
-        const nextValue = value instanceof Function ? value(curr) : value;
+        const nextValue = typeof value === 'function' ? value(curr) : value;
 
         if (nextValue == null) {
           storage.remove(key);
--- a/packages/react/react/src/hooks/useStorageState.ts
+++ b/packages/react/react/src/hooks/useStorageState.ts
@@ -2,36 +2,35 @@
 import { safeLocalStorage, Storage } from '@toss/storage';
 import { SetStateAction, useCallback, useState } from 'react';
 
-export type Serializable<T> = T extends string | number | boolean | unknown[] | Record<string, unknown> ? T : never;
-
+export type Serializable = string | number | boolean | unknown[] | Record<string, unknown>;
 interface StorageStateOptions<T> {
   storage?: Storage;
-  defaultValue?: Serializable<T>;
+  defaultValue?: T;
 }
 
 interface StorageStateOptionsWithDefaultValue<T> extends StorageStateOptions<T> {
-  defaultValue: Serializable<T>;
+  defaultValue: T;
 }
 
 /**
  * 스토리지에 상태값을 저장하고 불러와서 값이 보존되는 useState로 동작하는 hook입니다.
  * @param key 스토리지에 저장할 키
  */
-export function useStorageState<T>(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/react/react/src/hooks/useStorageState.ts [hunk 1]
```diff
@@ -2,36 +2,35 @@
...
 export function useStorageState<T>(
   key: string,
   { storage, defaultValue }: StorageStateOptions<T>
-): readonly [Serializable<T> | undefined, (value: SetStateAction<Serializable<T> | undefined>) => void, () => void];
+): readonly [T | undefined, (value: SetStateAction<T | undefined>) => void, () => void];
 export function useStorageState<T>(
   key: string,
   { storage = safeLocalStorage, defaultValue }: StorageStateOptions<T> = {}
-): readonly [Serializable<T> | undefined, (value: SetStateAction<Serializable<T> | undefined>) => void, () => void] {
+): readonly [T | undefined, (value: SetStateAction<T | undefined>) => void, () => void] {
```
ojj1123, line 33: @Jinho1011 
First of all, thank you for your interest in my PR.

Is there a reason for putting a generic constraint only in the first function definition? Could it be a mistake?

Jinho1011: Thank you for letting me know. I fixed the missing one and pushed a new commit with the changes.

raon0211:
Thanks for your contribution!

packages/react/react/src/hooks/useStorageState.ts
```diff
@@ -53,12 +52,12 @@ export function useStorageState<T>(
...
 
-  const [state, setState] = useState<Serializable<T> | undefined>(getValue);
+  const [state, setState] = useState<T | undefined>(getValue);
 
   const set = useCallback(
-    (value: SetStateAction<Serializable<T> | undefined>) => {
+    (value: SetStateAction<T | undefined>) => {
       setState(curr => {
-        const nextValue = typeof value === 'function' ? value(curr) : value;
+        const nextValue = value instanceof Function ? value(curr) : value;
```
raon0211, line 60: Is there a reason a `typeof` check is replaced with `instanceof` here?

Jinho1011: This change was caused by the mistake of generic constraint only on the first function among the overloading functions.

 If generic T is not constrained to extend serializable, T will not be callable, so I modified it like this. 

Since I fixed the generic constraint from the above conversation to be added to all overloaded functions, this change can be reversed.

raon0211: Thanks! What about reverting this change and making this pull request more focused on the main idea?

Jinho1011: I reverted the change as you said. Thank you for taking the time to review my pull request and provide feedback!

ojj1123:
@Jinho1011 
I was going to review it, but it has already been merged😭

Could you please re-review? @raon0211 

---

This PR didn't solve the issue #220 either:
type-error
(Please check my [typescript playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDKwoEsCGAbLALxwCM9UBeOAZxmwDsBzOAHznoFcBbEzVuEhAjkc9fh3oBrehADu9ANoBdfgCVgAY2gATADy0GjADRwJ0ufQB8AKGtZ6MTADMcG1GhjQcjYB5yOAeTAYLAh6al0AFUs4AG9rODhtYBcOPBgANXwOYAB+AC44SIBuawBfWycJDRCw02pfTyhvRv9gKLhQR3ptanRMXAJiMmBLAApYpJScNMzs1DLCjy8fP0Dg0PCoywBKQsjxHpT7YG14xKhgGA4oMWTU9Ky8HNKK6wB6d7gASScph7mz1QWD6ODgYGwXCwIQAbqgkCgTNC4CCUfQnJhLto4DhQXACI5mngEMhgAA6azkeCOWgARjgVA4DWWzVWMDaE3+M0e80KACIABbAPB4CB8uBlHalT6JAB6uUpVwQwFoACYGfVGitWo5OfduYCcoUACzqyXSr5weWK6kqmAAZg1TK1rJ1wD101mTyNCCgOQlUo+lutthlvy5XvmKLxMnoAFoIVgobD4aSTBCIChYFgVTjLuwIPB7BioFicX0CZh8CSUNQKVTlbRjU7mU0Wmt3ZN9ZGgYVJjhCrSTCRCgByAyjkwaQp0f1lAMWuUKhs0mAAVhbLvb7N1Xc9PN7cAUQ9VRntKnNQaXNsbMAAbJuWduOXuAd7gIVj0Zx3RJy48A0F6BjKVq5EAA))

In fact, I also tried with a `generic constraint`(you did) instead of a `conditional type constraint`(before).

If you insert a type constraint that prevents passing a `symbol` to `defaultValue`, the type parameter `T` is inevitably inferred as a literal type.
This is because `primitive types` (`string`, `number`, and `boolean`) are included in the `type constraint`. The reason is explained here #220.

I think that
**To prevent inference as a literal type, we need to remove the `type constraint` that prevents passing a `symbol` to `defaultValue`.
Otherwise, we can add a phrase that recommends writing `generic` to the official documentation:**
```ts
  const [state, setState, refresh] = useStorageState<number>('number', {
    defaultValue: 0,
  });
```

What do you think? @Jinho1011 @raon0211

Jinho1011:
@ojj1123 


I also confirmed that this PR did not solve the issue. I apologize for my mistake. But I found that the method below widens the type, but what is your opinion on its appearance, even if it is not pretty?


````typescript
type ToPrimitive<T> = T extends string
  ? string
  : T extends number
  ? number
  : T extends boolean
  ? boolean
  : T extends unknown[]
  ? unknown[]
  : T extends Record<string, unknown>
  ? Record<string, unknown>
  : T;
````

PR #234, opened 2023-04-28
fix(react-query): fix useSuspendedQuery to receive various argument formats like original react-query v3

manudeli:
# Fix useSuspendedQuery to receive various argument formats like original react-query v3

We can see below sentence in slash.page
> The basic API is the same as [react-query](https://tanstack.com/query/v4/?from=reactQueryV3&original=https://react-query-v3.tanstack.com/).

image

But, actually @toss/react-query's basic API is not same with react-query v3

@toss/react-query was still only available in the format below.
`arg1: queryKey, arg2: queryFn, arg3: Omit<SuspendedUseQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'queryKey' | 'queryFn'>`

But, anyway I thought we don't needed to change @toss/react-query's basic api to fit really well following react-query v3. Because it was not too inconvenient.

### Necessity to make change @toss/react-query
But recently, from @tanstack/react-query v5, I saw a [BREAKING CHANGE](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#supports-a-single-signature-one-object) guide that the form of parameters received in the future will not be received in various forms, but only in object form.
image

So, we need to make the queryKey of arg1 or the queryFn of arg2 possibly included in SuspendedUseQueryOptions.

## AS-IS
```tsx
const query = useSuspendedQuery(queryKey, queryFn, options)
```
## TO-BE
```tsx
// Now we can use the same function in these different ways.
const query = useSuspendedQuery(queryKey, queryFn, options)
const query = useSuspendedQuery(queryKey, { queryFn, ...options })
const query = useSuspendedQuery({ queryKey, queryFn, ...options })
```

So I created this Pull Request.

## Expected effect when merging this Pull Request
With this Pull Request, we can hope that @toss/react-query users will gradually migrate from react-query v3 to @tanstack/react-query v5 without big change

### Add [tsd](https://github.com/SamVerschueren/tsd) test case more to test d.ts files built
[tsd](https://github.com/SamVerschueren/tsd) test case was added more in this Pull Request to test that this method does not return the wrong type, so we can easily type-check the built d.ts while working.

### References
This Pull Request was made by referring to this Pull Request(https://github.com/suspensive/react/pull/38)

## PR Checklist


packages/react/react-query/src/utils/isQueryKey/index.ts
```diff
@@ -0,0 +1,5 @@
+import { QueryKey } from 'react-query';
+
+export function isQueryKey(value: any): value is QueryKey {
+  return typeof value === 'string' || Array.isArray(value);
+}
```
manudeli, line 5: https://github.com/TanStack/query/blob/v3/src/core/utils.ts#L423C17-L425

packages/react/react-query/src/hooks/useSuspendedQuery.test-d.ts
```diff
@@ -0,0 +1,73 @@
...
+expectType<undefined>(
+  useSuspendedQuery({
+    queryKey,
+    queryFn,
+    enabled: false,
+  }).data
+);
+
+// no arg
+expectError(useSuspendedQuery());
```
manudeli, line 73: Test all Type case. If there is another case, suggest it please

manudeli:
As much as there is a [@tanstack/eslint-plugin-query](https://tanstack.com/query/v4/docs/react/eslint/eslint-plugin-query)'s @tanstack/query/prefer-query-object-syntax rule, @tanstack/react-query tries to avoid receiving 2 or 3 arguments. Instead, it is induced by receiving only an object for one argument. Therefore, the current @toss/react-query must support receiving 2 or 1 arguments, the same as react-query v3.

raon0211 (approved):
This is great work. Thanks, @manudeli !

raon0211 (approved):
This is great work. Thanks, @manudeli !

raon0211 (approved):
This is great work. Thanks, @manudeli !

PR #236, opened 2023-05-06
refactor(@toss/react): `ClickArea` was converted class component to functional component. fixed docs.

ChanhyukPark-Tech:
## Overview

1. ClickArea was using class component rather than functional component 
 => The @toss/react library has a consistency . it usually uses functional components not class components

2. docs said `ClickArea` is wrapped a `div` tag. But it's actually an `anchor` tag.

## Question
Why is the `ClickArea` component in a way that gives the anchor tag a role(button) without using the button tag?

If it's not a situation where we can't use the button, I think it's better to use the standard button tag.

## PR Checklist


packages/react/react/src/components/ClickArea/ClickArea.tsx
```diff
@@ -14,30 +14,22 @@ interface Props {
...
-}
+export const ClickArea = ({ className, children, onClick, enabled = true }: Props) => {
+  return (
+    <Style css={css}>
+      <a
+        role="button"
+        className={classnames(className, CLASSNAMES.clickArea, { [CLASSNAMES.enabled]: enabled })}
+        // iOS에서 :active 선택자가 작동하지 않아 사용하는 Hack
+        // 참조: https://stackoverflow.com/questions/3885018/active-pseudo-class-doesnt-work-in-mobile-safari
+        onTouchStart={noop}
```
ChanhyukPark-Tech, line 25: I used @toss/utils's noop function

raon0211 (approved):
Thanks for refactoring!

PR #240, opened 2023-05-06
feat(@toss/validators): implemenet `isBirthDate8` function.

ChanhyukPark-Tech:
## Overview

It is related with [isBirthDate6](https://slash.page/libraries/common/validators/src/validators/is-birth-date-6.i18n) function

I am developing applications comfortably through good functions such as `isBirthDate6`. When I create an application, I may need to receive 8 digits of your birth date instead of 6 digits. 

So I created a new function. The internal implementation is almost the same.

## PR Checklist


HoseungJang (commented):
I think `isBirthDate8` is not proper to be added to `slash`, because a string representing a birthday could have various format.

In other words, the string of 8 characters is not a standard to represent a birthday, so `isBirthDate8` could not be common module.

ChanhyukPark-Tech:
@HoseungJang 
Thank you for the comments!
Where can I find the standard for birthday?
Is it an international standard to mark birthdays in six digits as isBirthDate6 validates? 😁

HoseungJang:
I think the purpose of `isBirthDate6` is close [to validate the front resident registration number of Korea](https://github.com/toss/slash/blob/f37ad540aa5aaa9712b06c91c369c12bf25f9efa/packages/common/validators/src/validators/is-rrn.ts#L9). And it's standard.

ChanhyukPark-Tech:
@HoseungJang Thank you for even showing examples of your own code!
I really understand

PR #241, opened 2023-05-23
fix(use-query-param): correct type of useQueryParam

manudeli:
## Overview
remove unnecessary generic of useQueryParam
1. if no options, return type of `useQueryParam` can be only `string | undefined`
2. if `useNextRouter.query['somthing']` is `string[]`, this case need to be avoid in this type implementation

## PR Checklist


packages/react/use-query-param/src/hooks/useQueryParam.ts
```diff
@@ -6,12 +6,16 @@ interface Options<T> {
...
+export function useQueryParam(name: string): string | undefined;
 export function useQueryParam<T = string>(name: string, options: Options<T>): T | undefined;
 export function useQueryParam<T = string>(name: string, options?: Options<T>) {
   const router = useNextRouter({ suspense: options?.suspense });
 
-  const value = router.query[name] as string | undefined;
+  const value = router.query[name];
+
+  if (Array.isArray(value)) {
+    throw new Error(`QueryParam of ${name} is Array, can't be returned by useQueryParam`);
```
HoseungJang, line 17:
```
`QueryParam of ${name} is Array, can't be returned by useQueryParam`
```

Why did you think so?

HoseungJang: And I think it cause an runtime exception when:

```typescript
const productIds = useQueryParam<string[]>('productIds');
```

HoseungJang: So I think the correct thing is to assert the return value as `T`.

manudeli:
@HoseungJang I fixed it all.

packages/react/use-funnel/src/useFunnel.tsx
```diff
@@ -56,13 +56,16 @@ export const useFunnel = <Steps extends NonEmptyArray<string>>(
     () =>
       Object.assign(
         function RouteFunnel(props: RouteFunnelProps<Steps>) {
-          // eslint-disable-next-line react-hooks/rules-of-hooks
-          const step = useQueryParam<Steps[number]>(stepQueryKey) ?? options?.initialStep;
+          const step = useQueryParam(stepQueryKey) ?? options?.initialStep;
```
manudeli, line 59: removed unnecessary type assertion in useFunnel

packages/react/use-funnel/src/useFunnel.tsx
```diff
@@ -56,13 +56,16 @@ export const useFunnel = <Steps extends NonEmptyArray<string>>(
...
 
           assert(
             step != null,
             `표시할 스텝을 ${stepQueryKey} 쿼리 파라미터에 지정해주세요. 쿼리 파라미터가 없을 때 초기 스텝을 렌더하려면 useFunnel의 두 번째 파라미터 options에 initialStep을 지정해주세요.`
           );
+          assert(
+            !Array.isArray(step),
+            `표시할 스텝을 ${stepQueryKey} 쿼리 파라미터에 배열타입으로 지정하면 안됩니다. string타입으로 지정해주세요.`
+          );
```
manudeli, line 68: step can be string[] type, so I did type narrowing with removing edgecase of queryparam

packages/react/use-query-param/src/hooks/useQueryParam.spec.ts
```diff
@@ -33,15 +33,19 @@ describe('useQueryParam', () => {
 
     mockUseRouter.mockReturnValue(router);
 
-    const { result, rerender } = renderHook(({ name, parser }) => useQueryParam(name, { parser }), {
-      initialProps: { name: 'foo', parser: Number as (val: string | string[]) => unknown },
```
manudeli, line 37: removed unnecessary type assertion

packages/react/use-query-param/src/hooks/useQueryParam.ts
```diff
@@ -1,17 +1,18 @@
...
-  parser?: (val: string) => T;
+interface Options<TParsed> {
+  parser?: (value: NextRouter['query'][string]) => TParsed;
   suspense?: boolean;
 }
 
-export function useQueryParam<T = string>(name: string): T | undefined;
-export function useQueryParam<T = string>(name: string, options: Options<T>): T | undefined;
-export function useQueryParam<T = string>(name: string, options?: Options<T>) {
+export function useQueryParam(name: string): NextRouter['query'][string];
```
manudeli, line 10: type correctly

packages/react/use-query-param/src/hooks/useQueryParam.ts
```diff
@@ -1,17 +1,18 @@
...
+export function useQueryParam<TParsed>(name: string, options: Options<TParsed>): TParsed;
```
manudeli, line 11: if options have parser field, ReturnType<typeof parser> is need to be return type of useQueryParam. so we can remove unnecessary undefined.

HoseungJang (commented):
The changes in this PR is not appropriate because `useQueryParam` is basically designed to return `string` when you don't pass any generic type. And It's designed for a single query parameter, not multiple query parameters like `?a=1&a=2&a=3` which are parsed to `string[]`.

And it is breaking changes [as you see](https://github.com/toss/slash/pull/241/files#diff-22b2ba85b7dc35c45d33851a75ae5751fa89db39d5776559c02b086baaa26168).

HoseungJang (commented):
The changes in this PR is not appropriate because `useQueryParam` is basically designed for a single query parameter, not an multiple query parameters like `?a=1&a=2&a=3` which are parsed to `['1', '2', '3']`.

And it is breaking changes [as you see](https://github.com/toss/slash/pull/241/files#diff-22b2ba85b7dc35c45d33851a75ae5751fa89db39d5776559c02b086baaa26168).

HoseungJang:
And I think `useQueryParam` is basically designed for a single query parameter like `?a=1`, not multiple query parameters like `?a=1&a=2&a=3` which are parsed to `string[]`.

HoseungJang:
These changes is not appropriate because I think `useQueryParam` is basically designed for a single query parameter like `?a=1`, not multiple query parameters like `?a=1&a=2&a=3` which are parsed to `string[]`.

Regardless that design is appropriate, it's difficult to merge this PR because it is breaking changes [as you see](https://github.com/toss/slash/pull/241/files#diff-22b2ba85b7dc35c45d33851a75ae5751fa89db39d5776559c02b086baaa26168).

manudeli:
> These changes are not appropriate because I think `useQueryParam` is basically designed for a single query parameter like `?a=1`, not multiple query parameters like `?a=1&a=2&a=3` which are parsed to `string[]`.
> 
> Regardless that design is appropriate, it's difficult to merge this PR because it is breaking changes [as you see](https://github.com/toss/slash/pull/241/files#diff-22b2ba85b7dc35c45d33851a75ae5751fa89db39d5776559c02b086baaa26168).

My reply is a bit late. It took me a while to figure out what the intentions were.

I checked the intent of `useQueryParam` hook in [code lines of useQueryParam.spec.ts](https://github.com/toss/slash/blob/main/packages/react/use-query-param/src/hooks/useQueryParam.spec.ts#L13-L25). The intent of this hook is to only use the first item of the selected queryParam. Then, even if the url is `https://example.com?item=1&item=2...`, it would be nice to guarantee it through a type guard so that it works as intended.

So I changed it like [32585b7](https://github.com/toss/slash/pull/241/commits/32585b7c3d6bd29cacdab2f384cf03a396ff7e63). Then, without breaking the design of this hook, we can get rid of unnecessary BREAKING CHANGE, type assertion, generic injection, and more stability in useFunnel on `https://example.com?item=1&item=2...`.

@HoseungJang, How do you think of this change? If not, could you suggest code more? I think you know what point I'm taking issue with this hook.

packages/react/use-funnel/src/useFunnel.tsx
HoseungJang: This is unnecessary change in this PR.

packages/react/use-query-param/src/hooks/useQueryParam.ts
```diff
@@ -1,21 +1,21 @@
...
+interface Options<TParsed> {
+  parser?: (value: Exclude<NextRouter['query'][string], unknown[]>) => TParsed;
   suspense?: boolean;
 }
 
-export function useQueryParam<T = string>(name: string): T | undefined;
-export function useQueryParam<T = string>(name: string, options: Options<T>): T | undefined;
-export function useQueryParam<T = string>(name: string, options?: Options<T>) {
+export function useQueryParam(name: string): Exclude<NextRouter['query'][string], unknown[]>;
+export function useQueryParam<TParsed>(name: string, options: Options<TParsed>): TParsed;
```
HoseungJang, line 11: I think this overload is essentially not necessary.
This is enough to be:
```typescript
export function useQueryParam<T = string>(name: string, options: Options<T> = {}): T | undefined {
  /* ... */
}
```

packages/react/use-query-param/src/hooks/useQueryParam.ts
```diff
@@ -1,21 +1,21 @@
 /** @tossdocs-ignore */
+import { NextRouter } from 'next/router';
 import { useNextRouter } from './useNextRouter';
 
-interface Options<T> {
-  parser?: (val: string) => T;
+interface Options<TParsed> {
+  parser?: (value: Exclude<NextRouter['query'][string], unknown[]>) => TParsed;
```
HoseungJang, line 6: Revert please.

packages/react/use-query-param/src/hooks/useQueryParam.ts
```diff
@@ -1,21 +1,21 @@
...
-  if (value == null || options?.parser == null) {
-    return value;
-  } else {
-    return options.parser(value);
-  }
+  return value == null || options?.parser == null
+    ? Array.isArray(value)
+      ? value[0]
+      : value
+    : options.parser(Array.isArray(value) ? value[0] : value);
```
HoseungJang, line 20:
```typescript
const value = (() => {
  const param = router.query[name];
  return Array.isArray(param) ? param[0] : param;
})();

if (value == null) {
  return undefined;
}

if (options.parser != null) {
  return options.parser(value);
}

return value;

```

HoseungJang (changes requested):

HoseungJang:
I have reviewed but I still think this change is not appropriate. Regardless the intention, someone has already used like:

```typescript
const _productIds = useQueryParam<string[]>('productIds');
const productIds = _productIds.map((id) => parseInt(id));
```

I understand that you want to contribute, but It would be better to keep the previous implementation.

manudeli:
> I have reviewed but I still think this change is not appropriate. Regardless of the intention of `useQueryParam`, someone may have already used it like:
> 
> ```ts
> const productIds = useQueryParam<string[]>('productIds');
> const parsedProductIds = productIds.map((id) => parseInt(id));
> ```
> 
> I understand that you want to contribute, but It would be better to keep the previous implementation.

Even in this case, there is an error when used with parser.
If url is `https://example.com?productIds=1&productIds=2`
```ts
const productIds = useQueryParam<string[]>('productIds', {
  parser: (productIds) => { // productIds will be ["1", "2"], but typely this is string... then, what role is parser?
    /* we can't use productIds.map((id) => parseInt(id)) */
  }
});
```
You can see example typely invalid example or runtimely invalid example
#### typely invalid (https://example.com?productIds=1&productIds=2)
image
#### runtimely invalid (https://example.com?productIds=1&productIds=2)
image


Even though there are clearly edge cases in the code, It is difficult to understand why Pull Requests or other code suggestion cannot be accepted only because it is a BREAKING CHANGE. So what is BREAKING CHANGE for and why do other libraries do major bumps? If necessary, I think we should proceed with a BREAKING CHANGE.

HoseungJang:
Why are you only obsessed by the parser option even in [my example](https://github.com/toss/slash/pull/241#issuecomment-1585073166)? It just says the case which could be affected by this breaking change.

A utility DO NOT HAVE TO completely support all of use cases. `useQueryParam` is the same.

It is a good utility that a utility that resolves many problems by a simple way.

Assume that `useQueryParam` supports the use of multiple query parameters too, the usage would be complicated.

I think it will be better and simpler to make a new hook(e.g. `useArrayQueryParam`) to support it.

PR #242, opened 2023-05-23
fix(use-query-param): do narrow type of useNextRouter

manudeli:
## do narrow type of useNextRouter
If using suspense, useNextRouter().isReady don't need to be boolean. so I corrected it and add test code.

### referenece
image

## PR Checklist

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -10216,6 +10216,7 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
             ["@types/react", "npm:18.0.21"],\
             ["@types/react-dom", "npm:18.0.6"],\
             ["@types/testing-library__jest-dom", "npm:5.14.5"],\
+            ["@yarnpkg/pnpify", "npm:4.0.0-rc.27"],\
             ["babel-jest", "virtual:785204e63ba96625a23f896ee431c9a36baff7a533742c7d2c1d3e8c51b41da4d9b22d61afa2e1bb44d0338b2a9e895a7c77e29d79f3a357a12dae28feb7f916#npm:29.1.2"],\
             ["concurrently", "npm:6.5.1"],\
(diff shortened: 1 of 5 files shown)


HoseungJang (commented):
I think this is too complicated. Because a person who set `suspense` to true basically doesn't have to check what the value `router.isReady` has.

manudeli:
> I think this is too complicated. Because if you set `suspense` to `true`, basically you don't have to check what the value `router.isReady` has.

@HoseungJang 
I think this Pull Request is meaningful because, like you said, router.isReady doesn't have to be checked. "Only if suspense: true". but how about, suspense: false or boolean?

1. Before this Pull Request, isReady came out as a boolean, so it may be confusing for users, but after this change, it is clear isReady: true.
2. if want to use useNextRouter as suspense: boolean, isReady must be boolean because we can't know whether to execute throw waitForRouterReady().
3. if want to use useNextRouter as suspense: false, isReady must be boolean because you can't execute throw waitForRouterReady().

It's not more complicated, I think it's more simpler. If you still think it's complicated, What's complicated? Some more explanation is needed.

HoseungJang:
> Before this Pull Request, isReady came out as a boolean, so it may be confusing for users, but after this change, it is clear isReady: true.

I think there isn't any confusing if users understand [how router.isReady will be changed.](https://nextjs.org/docs/pages/building-your-application/rendering/automatic-static-optimization)

> if want to use useNextRouter as suspense: boolean, isReady must be boolean because we can't know whether to execute throw waitForRouterReady().

I think it's helpful for you to see how suspense works.

And, to say once again, basically you don't have to check isReady when you set suspense to true, then how can this change be more simpler?

HoseungJang:
To be more clear, there are two use cases, checking `if (router.isReady)` or not.

Then, for the former case, is it meaningful to check the boolean value always evaluated as true?

Issue #243, opened 2023-05-29
[BUG]: typedStorage import type error


ChanhyukPark-Tech:
### **Package Scope**

Package name: @toss/storage

### **Describe the bug**

>TypedStorage is subpath exported module. You should use @toss/storage/typed to use it.

when I use that import statement type error occurs.
<img width="456" alt="image" src="https://github.com/toss/slash/assets/69495129/1c0ee757-ac95-40ad-87b2-3ccb8ef8ef9a">

because there is no typed file in node_modules
<img width="419" alt="image" src="https://github.com/toss/slash/assets/69495129/73a1fdbd-7482-440d-a891-1cebfea3fca6">

### **Expected behavior**

```tsx
import { TypedStorage } from "@toss/storage/typed"; // should not occur type error
```

### **To Reproduce**

I'm used classic yarn not yarn berry

[link](https://github.com/toppingskorea/mungther-survey/blob/main/src/apps/main/components/modals/SubmitModal.hooks.ts)

### **Possible Solution**

1. Using TypeScript paths option: Update the TypeScript configuration file (tsconfig.json) with the paths option to redirect module paths. By mapping @toss/storage/typed to @toss/storage/dist/typed, you can resolve the type error. Here's an example:

```tsx
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@toss/storage/typed": ["./node_modules/@toss/storage/dist/typed"]
    }
  }
}
```

but it is not good for DX
every project it should be considered

2. Directly move or copy files: Move or copy the contents of @toss/storage/typed to @toss/storage/dist/typed directly. By doing this, you can keep the import statement as it is without modifying it, and the type error should be resolved.

HoseungJang:
This fixed in https://github.com/toss/slash/pull/251.

And it's helpful for you to see [how module resolution work in Node.js](https://nodejs.org/api/packages.html), or [this article,](https://toss.tech/article/commonjs-esm-exports-field).

PR #247, opened 2023-06-03
fix(@toss/storage): update rollup-config for subpath export

minsoo-web:
## Overview

I Update `@toss/rollup-config` for `@toss/storage` package.
Because `@toss/storage/typed` is subpath module, but there is type error.

So I update `@toss/storage`\`s [**rollup.config.js**](https://github.com/toss/slash/commit/bbb4bc7a837e91a7baa8511fb135adf7b6b03a64) by using `@toss/rollup-config`, and [**update them**](https://github.com/toss/slash/commit/a31f9bcc38c3a10338076f08f17487b8861cb17f) cover the subpath module package.

### AS-IS

image

relates to #243 

### TO-BE

image

## PR Checklist


configs/rollup/package.json
```diff
@@ -1,14 +1,15 @@
 {
   "name": "@toss/rollup-config",
-  "version": "0.1.4",
+  "version": "0.1.5",
```
ChanhyukPark-Tech, line 3: you don't have to increase version manually
It will increased automatically by [ci](https://github.com/toss/slash/blob/main/.circleci/config.yml)

minsoo-web: Thx, I restore version to original [chore: cancel version changes in rollup-config](https://github.com/toss/slash/pull/247/commits/42c43e3864a4c20a31bda338846c2ec5dcbe4f72)

HoseungJang:
It is not the problem to be solved by fixing build script. It is the problem caused because you have not set `moduleResolution` to `nodenext` in `tsconfig.json`.

In other words, there are not any runtime errors because the problem is only caused in Typescript. (Currently Node.js resolves a module depending on `exports` field in `package.json`, please see [Node.js official documentation](https://nodejs.org/api/packages.html) or [this article](https://toss.tech/article/commonjs-esm-exports-field).)

HoseungJang:
So I think this can be fixed in very simple way:

1. Add `typed.d.ts` to the root.

```typescript
// typed.d.ts
export * from './dist/typed';
```

2. add `typed.d.ts` to `files` field in `package.json`.
```json
{
  "files": [
    "dist",
    "esm",
    "typed.d.ts",
  ]
}
```

HoseungJang:
I will close PR and fix the problem soon. Thanks.

minsoo-web:
Thank you @HoseungJang,
When I face this issue, find reference, I already read your post, but I finially realized that I didn`t understand about what is the problem.

If you don't mind, could you leave it to me to finish this matter?

Once again, thank you very much for reviewing my PR and for your kind response.

HoseungJang:
> If you don't mind, could you leave it to me to finish this matter?

Yes of course!

PR #250, opened 2023-06-03
test(@toss/react): wrote test code for usePreservedReference

rjsdnql123:
## Overview

A test code has been created to validate the behavior of the usePreservedReference.

## PR Checklist


packages/react/react/src/hooks/usePreservedReference.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+type TossObject = {
+  toss: string;
+};
+
+describe('usePreservedReference', () => {
+  const callbackReturnTrue = jest.fn(() => true);
+  const callbackReturnFalse = jest.fn(() => false);
+  const tossObject: TossObject = { toss: '토스' };
+
+  const useTestCustomHook = (object: TossObject, areValuesEqual?: () => boolean) => {
```
ChanhyukPark-Tech, line 15: type of `areValuesEqual ` in this file is `() => boolean`
but in implementation of usePreservedReference hook areValuesEqual type is `(a: T, b: T) => boolean`

Is there no problem if the two types are different?😄

ChanhyukPark-Tech (changes requested):

packages/react/react/src/hooks/usePreservedReference.spec.tsx
rjsdnql123: I thought it wouldn't be a problem!

The reason is that when `customFunction` is assigned to `areValuesEqual`, the expected behavior of `customFunction` is to return `true` if the **values are equal**, and `false` if the **values are different**.

Thank you for the review!😀

rjsdnql123: Thank you : )

packages/react/react/src/hooks/usePreservedReference.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const [reference, setReference] = useState<TossObject>(tossObject);
+    const ref = usePreservedReference<TossObject>(reference, areValuesEqual);
+
+    // 마운트 시 한번만 실행되는 코드
```
ChanhyukPark-Tech, line 19: It's not that complicated. I think the hook name shows the role well
```suggestion
```
Is there a reason you left an annotation?
If you need to leave an comment, you'd better leave it in English according to the convention😄

ssi02014:
https://github.com/toss/slash/pull/300 
I did the same thing without realizing the existence of that pull request.
However, the way we write our test code is slightly different.

I think it's possible to test usePreservedReference on its own without having to create a custom hook. It's a bit simpler and accomplishes our purpose.

Please take a look at the pull request and let me know what you think.

I don't care if my pull requests don't make it in.
I just want things to move in a better direction.

@ChanhyukPark-Tech @rjsdnql123

rjsdnql123:
I have reviewed @ssi02014 code, and I agree that using renderHook effectively makes it easier to maintain and more intuitive.

I believe submitting a pull request with @ssi02014 code seems like a good idea. 

what do you think

ssi02014:
@rjsdnql123
Yes, please edit your pull request.
I'm going to close the [#300](https://github.com/toss/slash/pull/300) pull request.

packages/react/react/src/hooks/usePreservedReference.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+    it('changed to the same value', () => {
+      const tossObject: TossObject = { toss: '토스' };
+
+      const { result, rerender } = renderHook(({ value }) => usePreservedReference(value), {
+        initialProps: { value: tossObject },
+      });
+
+      rerender({ value: { toss: '토스' } });
+
+      expect(tossObject).toBe(result.current);
```
ssi02014, line 22:
```suggestion
      expect(result.current).toBe(tossObject);
```
- Shouldn't the target of "expect" be "result.current"?

okinawaa (approved):
LGTM 👍

PR #259, opened 2023-06-14
test(@toss/utils): wrote test code for `getViewportSize` function

okinawaa:
## Overview

I wrote test code for getViewportSize function

## PR Checklist


packages/common/utils/src/getViewportSize.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    global.innerWidth = 1024;
+    global.innerHeight = 768;
+    const expectedSize = { width: 1024, height: 768 };
+    const size = getViewportSize();
+
+    expect(size).toEqual(expectedSize);
+  });
+
+  it('should return zero size when running on server', () => {
+    delete (global as { window: unknown }).window;
```
ChanhyukPark-Tech, line 14: I'd like to test the server status, should I fill it out like this?
Is it better to mock the `isServer` function itself

JayeHa: @okinawaa 
I think directly chaning the `window` object in your tests could be risky. 😢
I'm not sure that your tests won't affect the other tests.

How about mocking it like this instead?
I try it based on [this reference](https://claritydev.net/blog/mock-window-object-jest)!"

```js
import { getViewportSize } from '.';

let windowSpy: jest.SpyInstance;

beforeEach(() => {
  windowSpy = jest.spyOn(global, 'window', 'get');
});
afterEach(() => {
  windowSpy.mockRestore();
});

describe('getViewportSize', () => {
  it('should return viewport size', () => {
    windowSpy.mockImplementation(() => ({
      innerWidth: 1024,
      innerHeight: 768,
    }));

    const expectedSize = { width: 1024, height: 768 };
    const size = getViewportSize();

    expect(size).toEqual(expectedSize);
  });

  it('should return zero size when running on server', () => {
    windowSpy.mockImplementation(() => undefined);

    const expectedSize = { width: 0, height: 0 };
    const size = getViewportSize();

    expect(size).toEqual(expectedSize);
  });
});

```

raon0211 (approved):

Issue #257, opened 2023-06-11
[Feature]: Suggestion to add test code when using '이에/에' option in `josa` function


jeongtae:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: [@toss/hangul josa](https://slash.page/libraries/common/hangul/src/josa.i18n/)

### **Overview**

Hello there.
First of all, I've been using the useful hangul functions created by Toss, and I'd like to thank you for providing them.

I have been using the `josa` function, but I have noticed that the use of the `'이에/에'` option is a bit ambiguous.
So I read the [spec file](https://github.com/toss/slash/blob/4939042f2af98f610ded49620c844ebc4a3055f1/packages/common/hangul/src/josa.spec.ts), but it was missing a test case for that option.

The reason I can't find a use case for this option is probably due to my poor Korean skills, but if proper test cases are added, people like me can easily understand the use case.

### **Describe the solution you'd like**

Add a test case for the `'이에/에'` option to the test code of the `josa` function.

### **Additional context**

\-

PR #260, opened 2023-06-14
chore(@toss/hangul): delete not josa option '이에'

minsoo-web:
## Overview

I think `이에` option is useless option for  josa `에`

This is because `에` is used as `에` regardless of whether the previous word is `has받침` or `is종성ㄹ`

for example

```txt
저기에
책상에
지름길에
사이에 (is not case for word `사` but case for `사이`
```

So, I delete option `이에/에`

relates to  #257 

## PR Checklist


raon0211 (approved):
This is actually a breaking change but it seems that it is an incorrect option -- let's delete it.

PR #261, opened 2023-06-14
feat(@toss/hangul): add josa option with (으로서/로서)

minsoo-web:
## Overview

The `'로서'` josa is one of the spelling mistakes often mixed with `'로써' `
So, I add `'으로서/로서'` option for `josa` function

## Question

There is more josa option for example `'으로써/로써'`, `'으로부터/로부터'` ...etc
I wonder if josa function`s option list the more, the better, or if there is a standard.

## PR Checklist


raon0211 (approved):
Thanks!

PR #268, opened 2023-06-16
fix(error-boundary): fix useErrorBoundaryGroup to guarantee parent ErrorBoundaryGroup

manudeli:
## Fix useErrorBoundaryGroup to guarantee parent ErrorBoundaryGroup

I thought that when using useErrorBoundaryGroup, we need to guarantee parent ErrorBoundaryGroup to use itself.
So I added this Pull Request.

## PR Checklist


packages/react/error-boundary/src/ErrorBoundaryGroup.tsx
```diff
@@ -48,17 +48,29 @@ export const ErrorBoundaryGroup = ({
...
   return <ErrorBoundaryGroupContext.Provider value={value}>{children}</ErrorBoundaryGroupContext.Provider>;
 };
 
 export const useErrorBoundaryGroup = () => {
-  const { reset } = useContext(ErrorBoundaryGroupContext);
+  const group = useContext(ErrorBoundaryGroupContext);
+
+  if (group === undefined) {
+    throw new Error('useErrorBoundaryGroup: ErrorBoundaryGroup is required in parent');
+  }
```
minsoo-web, line 63: In [AsyncBoundaryContext](https://github.com/toss/slash/blob/6bb1eb5856c4ba231a9d418d2f477721be4aae63/packages/react/async-boundary/src/provider/AsyncBoundaryContext.tsx#L26), it is supposed to return the default value if there is no provider, so I wonder if it is a better way to generate an error. 

Do you think the method in AsyncBoundaryContext also needs to be improved?

manudeli: It'll be better in my opinion. And you?

I think before this Pull Request even If no parent boundary, this hook's returntype can make developer misunderstand as reset will work without Boundary. so I fixed it.

minsoo-web: > It'll be better in my opinion. And you?

That makes two of us. (I will make PR to AsyncBoundaryContext 🚀)

But, this hook`s return type, after this PR, also type like this

image

So, why don`t you add JSDoc for remindering to use Provider as parent.

As the `useContext` is like that

image

> as given by `the nearest context provider` for the given context.

manudeli: > But, this hook`s return type, after this PR, also type like this

Same type. but before this Pull Request, if no ErrorBoundaryGroup in parent, errorBoundaryGroup.reset will be () => {}, just empty function. I want to remove this case. If we use errorBoundaryGroup.reset function, I want to make developers to be confident that ErrorBoundaryGroup will be reset. so i did this.

anyway, Well informing developers with JSDoc seems like a good idea.

If this change is evaluate as not proper for slash maintainer, I want to suggest `undefined | {reset: () => void}` for ReturnType<typeof useErrorBoundaryGroup> to clarify that reset function can be useless when no ErrorBoundaryGroup in parent.

So first, I think we need to wait review first. when there is decision, I want to add jsdoc

minsoo-web: > I want to make developers to be confident that ErrorBoundaryGroup will be reset. so i did this.

I totally agree about this PR wants to do and agree throw error is better than just call empty default function called. 

Thank you for received my opinion 👍

packages/react/error-boundary/src/ErrorBoundaryGroup.tsx
```diff
@@ -48,17 +48,32 @@ export const ErrorBoundaryGroup = ({
...
+/**
+ * useErrorBoundaryGroup need ErrorBoundaryGroup in parent. if no ErrorBoundaryGroup, this hook will throw Error to prevent that case.
+ */
 export const useErrorBoundaryGroup = () => {
-  const { reset } = useContext(ErrorBoundaryGroupContext);
+  const group = useContext(ErrorBoundaryGroupContext);
 
-  return useMemo(() => ({ reset }), [reset]);
+  if (group === undefined) {
+    throw new Error('useErrorBoundaryGroup: ErrorBoundaryGroup is required in parent');
```
raon0211, line 65:
```suggestion
    throw new Error('useErrorBoundaryGroup must be used within an ErrorBoundaryGroup');
```

I guess this error message looks more natural in English to me -- what do you think?

packages/react/error-boundary/src/ErrorBoundaryGroup.tsx
raon0211: As you said, using `useErrorBoundaryGroup` where `ErrorBoundary` is not present is nonsensical. I think `throw`ing in advance makes more sense.

packages/react/error-boundary/src/ErrorBoundaryGroup.spec.tsx
```diff
@@ -54,7 +54,7 @@ describe('ErrorBoundaryGroup', () => {
 });
 
 describe('useErrorBoundaryGroup', () => {
-  it("useErrorBoundaryGroup's returning object is same with object returned in previous render", () => {
+  it('returning object is same with object returned in previous render', () => {
```
raon0211, line 57:
```suggestion
  it('returns the same object across renders', () => {
```

packages/react/error-boundary/src/ErrorBoundaryGroup.spec.tsx
```diff
@@ -65,4 +65,14 @@ describe('useErrorBoundaryGroup', () => {
 
     expect(prev).toBe(next);
   });
+
+  it('should throw error without ErrorBoundaryGroup in parent', () => {
```
raon0211, line 69:
```suggestion
  it('should throw error when it is used outside ErrorBoundaryGroup', () => {
```

packages/react/error-boundary/src/ErrorBoundaryGroup.spec.tsx
```diff
@@ -65,4 +65,14 @@ describe('useErrorBoundaryGroup', () => {
...
     expect(prev).toBe(next);
   });
+
+  it('should throw an error when it is used outside ErrorBoundaryGroup', () => {
+    const WithoutErrorBoundaryGroup = () => {
+      useErrorBoundaryGroup();
+      return <></>;
+    };
+    expect(() => render(<WithoutErrorBoundaryGroup />)).toThrowError(
+      'useErrorBoundaryGroup: ErrorBoundaryGroup is required in parent'
```
raon0211, line 75:
```suggestion
      'useErrorBoundaryGroup must be used within an ErrorBoundaryGroup'
```

raon0211 (approved):
Thanks, @manudeli ! This is a great work.

packages/react/error-boundary/src/ErrorBoundaryGroup.tsx
manudeli: > > It'll be better in my opinion. And you?
> 
> That makes two of us. (I will make PR to AsyncBoundaryContext 🚀 after this PR merged) It is better way of case for developer using this hook who don`t know what is wrong.
> 
> > this hook's returntype can make developer misunderstand
> 
> But, this hook`s return type, after this PR, also type like this
> 
> image
> So, why don`t you add JSDoc for remindering to use Provider as parent.
> 
> As the `useContext` is like that
> 
> image
> > as given by `the nearest context provider` for the given context.

@minsoo-web 
It's your time to make PR like this for AsyncBoundary as your comment. I just want to alert you.

PR #269, opened 2023-06-16
feat(@toss/utils): Add `ElementType` import from `@toss/utility-types`

minsoo-web:
## Overview

Change import path local to `@toss/utility-types`
Because this is likely to be a dualization of management.

## PR Checklist


ChanhyukPark-Tech:
This is redundant:)

ref https://github.com/toss/slash/pull/152#discussion_r1027409782

minsoo-web:
Thank you for your kind explanation.

> It could be a waste for people who don't use typescript!

I think this is different between `dependencies` and `devdependencies`.
I use `@toss/utility-types` for `devdependencies` that is not installed for other users (who install @toss/utils)

For example, if you install `@toss/react` that have typescript package in devdependencies.
But users node_modules is installed without typescript package includes any devdependencies. 

스크린샷 2023-06-18 오후 5 21 01

[reference. whats-the-difference-between-dependencies-devdependencies-and-peerdependencie](https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencie)

raon0211 (approved):
It was a duplicate -- thanks!

raon0211:
Actually, when using `ElementType` from `@toss/utility-types`, we should use it as a regular dependency (not a devDependency). Otherwise, the returned type from `pick` or `omit` would not be inferred. 

(Because `pick` returns `Pick<ObjectType, ElementType<KeyTypes>>`, and unless we install `@toss/utility-types`, the definition of `ElementType` is missing...)

raon0211:
This is a very tricky situation.

minsoo-web:
> and unless we install @toss/utility-types, the definition of ElementType is missing...)

Oh,,, I missed of the impact of an installed package on another package.

It seems that `@toss/utils` already provides a function of typed utility. (e.g pick, omit, object-values etc..)

My solution is to add more util types used in utils packages like [ObjectKeys](https://github.com/toss/slash/blob/6c3ef2f095c547a6d016c5fedca67f5235c1ee0b/packages/common/utils/src/object/object-keys.ts#L2) to `@toss/utility-types`.

I think it's the difference between providing **a lighter package** to the user or **guaranteeing the maintenance** of the `@toss/slash` package.

artechventure:
Adding "@toss/utily-types" to the dependencies is acceptable since it doesn't increase the size of the runtime bundle where it's utilized, despite potentially increasing the size of "@tossteam/utils" when fetched from npm.

minsoo-web:
> Adding "@toss/utily-types" to the dependencies is acceptable since it doesn't increase the size of the runtime bundle where it's utilized, despite potentially increasing the size of "@tossteam/utils" when fetched from npm.

Thank you for your kind explanation.

I have finished changing from devDependencies to dependencies. 
[chore: add @toss/utility-types to dependencies](https://github.com/toss/slash/pull/269/commits/cb96bf407307693382b86b62aaabb37bc4ecdac7)

raon0211 (approved):

raon0211:
Closing this pull request since this became stale. But we picked this change in the following commit: https://github.com/toss/slash/commit/063821639b8680bb1d0348306f43d26fb004a192. Thanks for your contribution!

PR #274, opened 2023-06-22
feat(@toss/react): add Portal Component

ssi02014:
## Overview

I worked on a Portal component to utilize the `createPortal` that React provides.

Please review and provide feedback on whether this component is useful and appropriate for "toss/slash". 🙏

For reference, I've added `@types/react-dom` to `@toss/react`.

```tsx
const Example = () => {
  return (
    <Portal>
      <Modal />
    </Portal>
  );
};
```

## PR Checklist


packages/react/react/src/components/Portal/index.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const outer = container.querySelector('#outer');
+
+    const parentChildren = parent?.querySelector('#child');
+    const outerChildren = outer?.querySelector('#child');
+
+    expect(parentChildren).not.toBeInTheDocument();
+    expect(outerChildren).toBeInTheDocument();
+  });
+
+  it("If the ID of a DOM Node added outside of the parent component's DOM hierarchy is different from the ID added with the Portal Component props, the Portal Component's Children are rendered as children of the parent component.", () => {
```
ChanhyukPark-Tech, line 31: I think many users who use `Portal` component do now want this feature😄
How about writing this part in the `document` because I think it's a side effect

packages/react/react/src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,22 @@
...
+/** @tossdocs-ignore */
+export function Portal({ id, children }: Props) {
+  const isMount = useIsMounted();
+
+  /**
+   * With this code, it is possible to solve the "window is not defined" and "Hydration Error" that can occur in SSR.
+   */
+  if (!isMount) {
+    return <></>;
+  }
```
ChanhyukPark-Tech, line 18:
```suggestion
  const isMounted = useIsMounted();

  /**
   * With this code, it is possible to solve the "window is not defined" and "Hydration Error" that can occur in SSR.
   */
  if (!isMounted) {
    return <></>;
  }
```

I think this is more appropriate😄

### Usage

image

ssi02014: That's something I missed..!!!

Thank you for the feedback. I agree with your opinion 👍

I'll fix that code in a moment🙆‍♂️

packages/react/react/src/components/Portal/index.spec.tsx
ssi02014: I understood your comment as adding what you mentioned in the ".md" document. 

Am I understanding your comment correctly?!

If this is correct, I will fix this as well!

Thank you for the feedback 🙂🙂

ChanhyukPark-Tech: yes I think it is good to be in .md file 😄

ssi02014:
@ChanhyukPark-Tech 
Fixed it! 🙏

packages/react/react/src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,22 @@
+import ReactDOM from 'react-dom';
+import { useIsMounted } from '../../hooks/useIsMounted';
+
+interface Props {
+  children: React.ReactNode;
+  id: string;
```
ChanhyukPark-Tech, line 6: How about trying to find the parent's node to render using the `ref` value instead of the `id`(string)?

Looking at the various implemented libraries, it seems to be implemented as `ref`, not as the string value id.

If it is implemented as an `id`, which is a normal magic string, it is difficult to find which file has the element with the `that id`, and it is likely to cause side effects.

However, if you manage it with `ref`, the location of the parent node can be directly identified on the code, so I think the developer experience will be good 

ref

[chakra-ui](https://chakra-ui.com/docs/components/portal/props)
[mui](https://mui.com/base-ui/react-portal/components-api/)

ssi02014: Thanks for the great feedback. 😀

Definitely, managing by `id(string)`  is more likely to cause side effects, and it's a bad developer experience. 

I will think about a Portal Component managed by `ref`. 

Additionally, the default behavior of "mui" is to add a span tag to the "body" tag, and "chakra ui" is to add a div tag with "chakra-portal" as class to the body tag.

The default behavior of chakra ui seems more appropriate to me, and I will try to design similarly to it.

ssi02014: @ChanhyukPark-Tech 
Oh sorry, I was mistaken about the MUI default behavior.

ssi02014:
@ChanhyukPark-Tech 
I agree with you and have modified the overall structure and functionality of the "Portal" component. 😀

The Portal component now renders the Portal Node in `document.body` by default.

However, if the user passes in the `containerRef` prop, it is possible to render the Portal Node in a DON Node other than document.body. 

Additionally, it supports nested portals. (But, I don't think nesting is the primary purpose of this component.)

Thank you again for the great feedback 🙏

packages/react/react/src/components/Portal/Portal.en.md
```diff
@@ -0,0 +1,77 @@
...
+[React CreatePortal](https://react.dev/reference/react-dom/createPortal#rendering-a-modal-dialog-with-a-portal) Please refer to the following articles
+
+<br />
+
+```tsx
+function Portal({
+  children,
+  containerRef,
+}: {
+  children: React.ReactNode;=
```
ChanhyukPark-Tech, line 20:
```suggestion
  children: React.ReactNode;
```

ssi02014: @ChanhyukPark-Tech 
OMG.. Thank you. I'll fix it right now

packages/react/react/src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const PORTAL_DEFAULT_CLASS = 'portal';
+
+function RenderPortal({ children, className, containerRef }: PortalProps) {
+  const { parentPortal } = useContext(portalContext);
+
+  const getPortalNode = useCallback(
+    (mountNode: HTMLElement) => {
+      const portalNode = mountNode.ownerDocument.createElement('div');
+      portalNode.classList.add(className || PORTAL_DEFAULT_CLASS);
```
ChanhyukPark-Tech, line 23: What do you think is better?
If you're going to set it as the default anyway, how about setting it from the beginning?

```suggestion
function RenderPortal({ children, className = PORTAL_DEFAULT_CLASS, containerRef }: PortalProps) {
  const { parentPortal } = useContext(portalContext);

  const getPortalNode = useCallback(
    (mountNode: HTMLElement) => {
      const portalNode = mountNode.ownerDocument.createElement('div');
      portalNode.classList.add(className);
```

packages/react/react/src/components/Portal/Portal.en.md
ChanhyukPark-Tech: You can easily fix it just by pressing the `commit suggestion` button on my comment!

ssi02014: @ChanhyukPark-Tech 
.....😂😂😂😂😂😂😂😂😂
Thank you so much for these tip. I was being an idiot and modifying the code myself every time, registering commits one by one.

packages/react/react/src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+/** @tossdocs-ignore */
+export function Portal({ children, ...restProps }: PortalProps) {
+  const isMounted = useIsMounted();
+
+  /**
+   * With this code, it is possible to solve the "window is not defined" and "Hydration Error" that can occur in SSR.
+   */
+  if (!isMounted) {
+    return <></>;
```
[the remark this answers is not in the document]
ssi02014: @ChanhyukPark-Tech 
My intention is simply that Portal's return type is inferred to be 'JSX.Element | null'. 

If you use Fragment, it will be inferred as 'JSX.Element' instead of 'JSX.Element | null'.

But neither is a big deal functionally... 👍

ChanhyukPark-Tech (approved):
thanks @ssi02014 

I think this component is good.
other slash maintainer, please double check.

packages/react/react/src/components/Portal/Portal.tsx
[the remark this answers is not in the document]
manudeli: neither is a big deal functionally... but Isn't it better to express it clearly to user: `this component can return nothing` by using null?

https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null

I could understand your intention, but I want to know why It's better `JSX.Element` than `JSX.Element | null`.
because I thought it's more clear,`JSX.Element | null`. I want to know case that why you need just `JSX.Element`. (I'm not sure if this is correct, so I'm curious about what many people think. 
@ssi02014, @ChanhyukPark-Tech, @raon0211, @HoseungJang etc.)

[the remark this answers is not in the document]
ssi02014: @manudeli
Thanks for the feedback. 🙏

I too was initially torn between null and Fragment.

My initial intent was simply that the Portal would render based on whether it was mounted or not, so I assumed that it would always render unless there were special circumstances, such as an error.

So, I felt it was awkward to have null as a return type, and since neither null nor Fragment were functionally problematic, I chose Fragment.

However, I'm going to defer to the majority here because I don't think this debate is a functional issue, but a matter of preference. 🙇‍♂️

PR #276, opened 2023-06-23
feat(@toss/react): add useLazyImage

ssi02014:
## Overview
I worked on the `useLazyImage` hook, which utilizes the `Intersection Observer`.

A hook that loads an image when the image tag given by `ref` is visible in the browser viewport (or the element specified by root).

Given as a return value, `isLoading` allows you to verify that the image has finished loading.

By applying Image Lazy Load, users no longer need to load all image resources at once when entering the page, improving their user experience.

Please review and provide feedback on whether this hook is useful and appropriate for "toss/slash". 🙏

## Demo Video
https://github.com/toss/slash/assets/64779472/bf4a9f01-58b7-4d67-9a1d-90812e9bba71

## Demo Code
```tsx
import imageUrl from './assets/img.png';

const Example = () => {
  const { ref, isLoading } = useLazyImage({ src: imageUrl });

  return (
    <div>
      {isLoading && <div>image is being loading...</div>}
      <img
        ref={imgRef}
        width={400}
        height={400}
        style={{ opacity: isLoading ? 0 : 1, transition: 'opacity 0.2s' }}
      />
    </div>
  );
};
```

## PR Checklist


packages/react/react/src/hooks/useImageLazyLoading.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    if (element.tagName !== 'IMG') {
+      throw new Error("The target element can only be an 'img' tag.");
+    }
+
+    // Checks if undefined or null.
+    if (src == null) {
+      throw new Error('The useImageLazyLoading hook requires the src property as a required value.');
+    }
```
HoseungJang, line 31: These are not necessary because I think we only need to consider Typescript developers.

packages/react/react/src/hooks/useImageLazyLoading.ts
```diff
@@ -0,0 +1,70 @@
...
+      root,
+      rootMargin,
+      threshold,
+    });
+
+    observer.observe(element);
+
+    return () => {
+      observer.unobserve(element);
+    };
```
HoseungJang, line 66: It looks good but how about using `useImpressionRef` in `@toss/impression-area`?

HoseungJang:
This looks good but there's no way to know when a image is loaded, when a element appears in viewport, etc.

I think there're some use cases like:
- Show a skeleton during loading an image.
- Apply fade-in animation to show an image when loading complete.

packages/react/react/src/hooks/useImageLazyLoading.ts
ssi02014: At first, I also considered using `useImpressionRef`, but since there were parts that did not match my intention 100%, such as unobserving the img tag, I rewrote the code. 🙇‍♂️

packages/react/react/src/hooks/useImageLazyLoading.ts
```diff
@@ -0,0 +1,70 @@
...
+interface Props {
+  src: string;
+  threshold?: number | number[];
+  root?: Document | Element | null;
+  rootMargin?: string;
+  onAction?: () => void;
+}
+
+/** @tossdocs-ignore */
+export function useImageLazyLoading<Element extends HTMLImageElement>({
```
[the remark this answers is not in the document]
ssi02014: Thank you for the feedback. I agree with your opinion 👍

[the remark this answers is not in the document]
ssi02014: Thank you for the feedback. Upon further review, I realized that I am using the `useRefEffect` hook and returning the `ref` of that hook, so it would be appropriate to change the return type to `EffectRef<HTMLImageElement>`.

packages/react/react/src/hooks/useImageLazyLoading.ts
```diff
@@ -0,0 +1,70 @@
...
+  src,
+  rootMargin,
+  threshold,
+  root,
+  onAction,
+}: Props) {
+  const ref = useRefEffect<Element>(element => {
+    if (typeof IntersectionObserver === 'undefined') {
+      return;
+    }
```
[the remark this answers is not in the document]
ssi02014: Thank you for the feedback. I agree with your opinion 👍

ssi02014:
@HoseungJang 
Thank you for the feedback!! 🙇‍♂️🙇‍♂️ 
I'd like to respond to a few things, so I'll do so in the comments!

When designing this hook for the first time, I took into account the `loading="lazy"` option for `img` tags and worked on it.

However, some browsers don't support `loading="lazy"`, so I've worked a hook that adds basic functionality and the ability to take additional actions when exposed to the viewport (or root).

docs: https://web.dev/browser-level-image-lazy-loading/

For the first case you mentioned, I think it is a conditional rendering case where the skeleton is shown while the API is being called and in loading state, and then the img tag is shown when the API call is completed. Is that correct?

If this is the case, we shouldn't have much of a problem because once the API call is complete and the img tag is rendered, we'll load the image when we need it

https://github.com/toss/slash/assets/64779472/4046d24d-9715-4f7a-8910-f264d39840ce

If you're applying animation after it's been exposed on the viewport (or root), this hook may not be suitable. If the image size is large, it may take a long time to load the image.

Additionally, the ultimate purpose of this hook is to prevent loading large amounts of images during initial rendering and to improve rendering performance.

Thank you for considering these purposes 🙇‍♂️

HoseungJang:
@ssi02014 I think I miscommunicated my opinion.

> For the first case you mentioned, I think it is a conditional rendering case where the skeleton is shown while the API is being called and in loading state, and then the img tag is shown when the API call is completed. Is that correct?

No, the skeleton what I said is not related to API fetching. I mean, it is shown when a browser is loading an image.

> If you're applying animation after it's been exposed on the viewport (or root), this hook may not be suitable. If the image size is large, it may take a long time to load the image.

I don't mean that the animation is applied right after the image appears in viewport. I mean, the animation is applied after the image is fully loaded.

I think it is easier to understand what I mean with an example:

```tsx
const [isLoading, setIsLoading] = useState(true);

return (
  <div>
    {isLoading && <div>image is being loading...</div>}
    <img
      src={imageURL}
      alt=""
      onLoad={() => setIsLoading(false)}
      style={{ opacity: isLoading ? 0 : 1, transition: 'opacity 0.2s' }}
    />
  </div>
);
```

1. First, `isLoading` is true because the image is being loading.
2. Show the skeleton `<div>image is being loading...</div>` if `isLoading` is true.
3. Hide the image by setting `opacity` to 0 of the image, when `isLoading` is true.
4. Set `isLoading` to false when the image is fully loaded.
5. Then, the skeleton is hidden and `opacity` of the image is set to 1.
6. So the image is faded in by `transition: opacity 0.2s`.

Such case is what I said.

ssi02014:
@HoseungJang 
Oh! I was completely thinking wrong....🥲

Thank you for the kind explanation.

After thinking a little more about how to improve that part, I will leave another reply. Thank you!

HoseungJang:
And I think there're more advantages to implement `useLazyImage` as React component.

1. Using hook tends to make the code verbose.
2. In this case I think developers can easily migrate if we provides it as React component.

Assume that an original code looks like:
```tsx
return (
  <>
    <img src={imageURL1} alt="" width={200} height={200} />
    <img src={imageURL2} alt="" width={200} height={200} />
    <img src={imageURL3} alt="" width={200} height={200} />
  </>
);
```

First, if we provides this feature as React hook:
```tsx
import { useLazyImage } from '@toss/react';

const imageRef1 = useLazyImage({ src: imageURL1 });
const imageRef2 = useLazyImage({ src: imageURL2 });
const imageRef3 = useLazyImage({ src: imageURL3 });

return (
  <>
    <img ref={imageRef1} alt="" width={200} height={200} />
    <img ref={imageRef2} alt="" width={200} height={200} />
    <img ref={imageRef3} alt="" width={200} height={200} />
  </>
);
```

And second, If we provides it as React component:
```tsx
import { LazyImage } from '@toss/react';

return (
  <>
    <LazyImage src={imageURL1} alt="" width={200} height={200} />
    <LazyImage src={imageURL2} alt="" width={200} height={200} />
    <LazyImage src={imageURL3} alt="" width={200} height={200} />
  </>
);
```

I think the latter example looks better.

ssi02014:
@HoseungJang 

I also agree with this opinion, it's very good! 👍

I was actually considering a "LazyImage" component as well, if the functionality of "useLazyHook" was implemented properly and reflected safely.

ssi02014:
@HoseungJang 
wow! 🤩

After testing, we found that the `useLazyImage` hook was working correctly as intended with the `onLoad` function!!!

Through this test, I also found out that this feature is working properly for the first time.

It's late at night, but I'm leaving a comment because I want to share the news 🙇‍♂️🙇‍♂️🙇‍♂️

Thank you!

## Test Video
https://github.com/toss/slash/assets/64779472/6a17fa48-bf7f-4923-9944-82360519dcf6

## Example Code
```jsx
const Example = () => {
  const [isLoading, setIsLoading] = useState(true);
  const ref = useLazyImage({ src: png1 });
  // ...

  return (
    <div>
      <Box>
        {isLoading && <Skeleton>로딩중</Skeleton>}
        <img
          ref={ref}
          width={400}
          height={400}
          onLoad={() => setIsLoading(false)}
          style={{ opacity: isLoading ? 0 : 1, transition: 'opacity 0.2s' }}
        />
      </Box>
      {/* ... */}
    </div>
  )
}
```

HoseungJang:
> My guess is that after an "img" is exposed to the "viewport" (or root) via the "Intersection Observer", the src attribute value of that img tag is assigned and "onLoad" is called normally once the image has finished loading.

Yes, `onLoad` called when `load` event fires, which called when an image is fully loaded.

[HTMLElement: load event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/load_event)

HoseungJang:
You can use `Image` class:

```typescript
const [src, setSrc] = useState(undefined);

useEffect(() => {
  const image = new Image();
  image.src = imageURL;

  // https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete
  // If the image had already been loaded and cached in a browser,
  // the initial value of complete property is true.
  if (image.complete) {
    setSrc(imageURL);
    return;
  }

  image.onload = () => {
    setSrc(imageURL);
  };
}, []);

return <img src={src} />;
```

ssi02014:
Good!
I already understood how "onLoad" works, but I think I was confused by the "intersection observer" in the middle. 🫠

As a result, I think it's safe to conclude that these issues are not a problem, but what do you think? 

> Show a skeleton during loading an image.
> Apply fade-in animation to show an image when loading complete.

HoseungJang:
> As a result, I don't think the following should be a problem. what do you think?

I think it's important to locate related code in a single point. I think because the place starts loading an image is `useLazyImage`, so the place checks loading completes should be `useLazyImage`.

ssi02014:
@HoseungJang 
A-ha!  I understand what you're saying.
I'll think about that and write a comment.
Thanks for the feedback!

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,67 @@
...
+      rootMargin,
+      threshold,
+    });
+
+    observer.observe(element);
+
+    return () => {
+      observer.unobserve(element);
+    };
+  }, []);
```
HoseungJang, line 64: I think deps should be added.
- src
- root
- onInView
- ...

I recommend you not to use `useRefEffect` because eslint error is ignored. Use `useEffect` instead:
```typescript
const ref = useRef<HTMLImageElement>(null);

useEffect(() => {
  const image = ref.current;
  if (image == null) {
    return;
  }

  image.src = src;
  // ...
}, [src, /* ... */]);
```

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,67 @@
+import { useState } from 'react';
+import { useRefEffect } from './useRefEffect';
+
+interface Props {
+  src: string;
+  threshold?: number | number[];
+  root?: Document | Element | null;
+  rootMargin?: string;
+  onInView?: () => void;
+}
```
HoseungJang, line 10: How about adding `onLoad` prop too?

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/** @tossdocs-ignore */
+export function useLazyImage({ src, rootMargin, threshold, root, onInView }: Props) {
+  const [isLoading, setIsLoading] = useState<boolean>(true);
```
HoseungJang, line 14: I think the initial value of this state should be set to `false` because image loading starts when an element appears in viewport. Then:

```typescript
const imgElement = entry.target as HTMLImageElement;
imgElement.src = src;

setIsLoading(true);
imgElement.onload = () => {
  onLoad();
  setIsLoading(false);
};
```

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const ref = useRefEffect<HTMLImageElement>(element => {
+    if (typeof IntersectionObserver === 'undefined') {
+      element.src = src;
```
HoseungJang, line 18: `isLoading` should be set in this block.

ssi02014:
@HoseungJang 
I made the following changes to the `useLazyImage` hook. 
After changing the code for `useLazyImage`, I tested it and found that it worked fine 👍 

```tsx
export function useLazyImage({ src, rootMargin, threshold, root, onInView }: Props) {
  const [isLoading, setIsLoading] = useState<boolean>(true); // (*)

  const ref = useRefEffect<HTMLImageElement>(element => {
    // ...
    const insertImageSrc = ([entry]: IntersectionObserverEntry[], observer: IntersectionObserver) => {
      if (entry) {
        if (entry.isIntersecting) {
          const imgElement = entry.target as HTMLImageElement;
          imgElement.src = src;
          
          // (*)
          imgElement.onload = () => {
            setIsLoading(false);
          };

          if (onInView) {
            onInView();
          }
          observer.unobserve(entry.target);
        }
      }
    };
    // ...
  }, []);

  return { ref, isLoading } as const; // (*)
} 
```
## Example Code
```tsx
const Example = () => {
  const { ref, isLoading } = useLazyImage({ src: ImageURL });

  return (
    <div>
      {isLoading && <Skeleton>로딩중</Skeleton>}
      <img
        ref={ref}
        width={400}
        height={400}
        style={{ opacity: isLoading ? 0 : 1, transition: 'opacity 0.5s' }}
      />
    </div>
  );
```
## Video
https://github.com/toss/slash/assets/64779472/72ea0c2c-9764-4c6b-9128-2af621b3f571

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,67 @@
...
+      return;
+    }
+
+    const insertImageSrc = ([entry]: IntersectionObserverEntry[], observer: IntersectionObserver) => {
+      if (entry) {
+        if (entry.isIntersecting) {
+          const imgElement = entry.target as HTMLImageElement;
+          imgElement.src = src;
+
+          imgElement.onload = () => {
```
HoseungJang, line 33: If `imgElement.complete` is true, we don't have to register `onload` handler. Because it means an image has already been loaded and cached in browser memory.

```typescript
if (imgElement.complete) {
  /* ... */
} else {
  imgElement.onload = () => {
    /* ... */
  };
}
```

packages/react/react/src/hooks/useLazyImage.ts
ssi02014: I get it. I'll fix that.

packages/react/react/src/hooks/useLazyImage.ts
ssi02014: Is it correct that the onLoad function refers to the function that is executed when the image finishes loading?

packages/react/react/src/hooks/useLazyImage.ts
ssi02014: I missed it... Thanks for the detailed review.

packages/react/react/src/hooks/useLazyImage.ts
HoseungJang: Exactly. But it can be subdivided:

- onLoadStart
- onLoadComplete

But I think it's not necessary now, it looks good to add only `onLoad`.

packages/react/react/src/hooks/useLazyImage.ts
ssi02014: @HoseungJang 
I've tested it, and the moment an `imgElement` is exposed to the viewport using the `intersection observer` and the `src` attribute is assigned, whether it's cached in the browser or not, the `complete` attribute becomes false.

If you don't use "imgElement", but instead use "new Image()", you get the same result. 

So, I don't know about other cases, but for the `useLazyImage` hook that utilizes the `intersection observer`, I don't think the logic to check `complete` is necessary.

HoseungJang: In Safari I didn't encounter the problem. It works as I said. I think there is a bug in Chrome.

ssi02014: @HoseungJang 
As you said, it doesn't work correctly only in chrome. 
I have confirmed that it works fine in safari and firefox.

ssi02014:
@HoseungJang 
I've made some changes to the overall structure of the `useLazyImage` hook based on feedback.
I know you can be annoying, but I really appreciate your continued feedback.

packages/react/react/src/hooks/useLazyImage.ts
ssi02014: I added "onLoadComplete" because "onLoad" can be confusing with "onLoad" in the img tag.

packages/react/react/src/hooks/useLazyImage.ts
ssi02014: Oh, that's right. I've added that code. Thanks. 
>And I think there is no reason not to add this logic even though complete property doesn't work correctly (only) in Chrome.

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,111 @@
...
+    (element: HTMLImageElement) => {
+      setIsLoading(true);
+
+      element.onload = () => {
+        setIsLoading(false);
+        onLoadComplete();
+      };
+    },
+    [onLoadComplete]
+  );
```
HoseungJang, line 40: I think this is closer to 'register' an event handler, than to be called when an event fires. So I think `registerLoadEventHandler` is more correct name.

ssi02014: @HoseungJang 
Oh... That's right

I think your suggestion is more appropriate since it involves registering an "onload" event handler on the img element.

I'll fix that code.

packages/react/react/src/hooks/useLazyImage.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (typeof IntersectionObserver === 'undefined') {
+      insertImageSrc(imgElement);
+      return;
+    }
+
+    if (imgElement.getAttribute('src')) {
+      console.warn('If the "src" attribute is initially in an "img" tag, lazy load is not applied.');
+      registerLoadEventHandler(imgElement);
+      return;
+    }
```
HoseungJang, line 95: I think this is an unnecessary guard because it could cause confusing.

If I pass an `src` 'A' to `useLazyImage`, I think I want it to load 'A', whatever `src` an element already has is.

Since the role of `useLazyImage` is to load an given image lazily, I think it should do its own role.

```typescript
const { ref } = useLazyImage({ src: 'A' });

return <img ref={ref} src='B' />
```

HoseungJang: And I think this kind of confusing essentially comes from the role of `useLazyImage`, to inject a `src` attribute to the image element.

I think we could make the role lighter and the code more intuitive like:
```tsx
const { ref, inView } = useInView({ root, rootMargin, threshold });
const { isLoaded } = useImageLoad(imageURL, { disabled: !inView });

return (
  <div>
    <div ref={ref} />
    {!isLoaded ? (
      <div style={{ width: 300, height: 300 }} /> // skeleton
    ) : (
      <img src={imageURL} style={{ width: 300, height: 300 }} />
    )}
  </div>
);
```

HoseungJang: Slash also provides [ImpressionArea](https://github.com/toss/slash/blob/fe40c2942fd1b4ddbf1dc9b34dffdffa2b91ccae/packages/react/impression-area/src/ImpressionArea.tsx#L31) and [useImpressionArea](https://github.com/toss/slash/blob/main/packages/react/impression-area/src/useImpressionRef.tsx).

So I think the most ideal situation is that we provides smaller modules as much as possible, and developers use them depending on what they need.

```tsx
const [inView, setInView] = useState(false);
const { isLoaded } = useImageLoad(imageURL, { disabled: !inView });

return (
  <ImpressionArea onImpressionStart={() => setInView(true)} onImpressionEnd={() => setInView(false)}>
    {!isLoaded ? (
      <div style={{ width: 300, height: 300 }} /> // skeleton
    ) : (
      <img src={imageURL} style={{ width: 300, height: 300 }} />
    )}
  </ImpressionArea>
);
```

HoseungJang: And another reason to make `useLazyImage` lighter is that it is unintuitive to start loading the image when the element appears in viewport. Actually it is implicit in the naming.

ssi02014: @HoseungJang 

>I think this is an unnecessary guard because it could cause confusing.
If I pass an src 'A' to useLazyImage, I think I want it to load 'A', whatever src an element already has is.
Since the role of useLazyImage is to load a given image lazily, I think it should do its own role.

I agree with you. This situation can happen enough. 🥲
That's why I wrote a minimalistic guard code to not put images directly into the src.
I thought that code would prevent this mess.

ssi02014: @HoseungJang

I was thinking that the "useLazyLoad" hook, abstracted and designed to be specific to "image" and "lazy load", was intuitive and good for usability.

But you're also right, ideally we'd like to provide smaller, more modular hooks and let the user pick and choose which ones to use.

Additionally, "ImpressionArea" is constantly watching its element. 

From an image perspective, once an image is loaded, it doesn't need to be observed anymore. From this perspective, "ImpressionArea" is not suitable at this point in time. Additional functionality is needed to make it work for this purpose.

For example, I think we need a "disabled" option to mean that we don't want to observe it anymore.

(Since "useInView" overlaps with "ImpressionArea" in many ways, we believe it would be better to improve "ImpressionArea".)

ssi02014: @HoseungJang 
I don't feel that the current issues with "useLazyImage" are critical, and I've modified the code several times based on your feedback and feel it's pretty good.  

But I also agree with you. Since you're the maintainer of that repository, I'm going to go with what you think.

Keeping that pull request open for too long is not a good direction.🙇‍♂️

HoseungJang: > (Since "useInView" overlaps with "ImpressionArea" in many ways, we believe it would be better to improve "ImpressionArea".)

`useInView` was just an example.

> Additionally, "ImpressionArea" is constantly watching its element.
>
> From an image perspective, once an image is loaded, it doesn't need to be observed anymore. From this perspective, "ImpressionArea" is not suitable at this point in time. Additional functionality is needed to make it work for this purpose.

I think this is related to making a feature to unobserve Intersection Observer anytime, and not related to what is the role `useLazyLoad` should have.

And although an observer handler is called every time, I think it would not have too bad impact on performance. It would be important not to re-render every time:

```tsx
const [inView, appear] = useBooleanState(false);

// No re-rendering will be triggered after `appear` is called first time. (`inView` will be always true)
return <ImpressionArea onImpressionStart={appear} />;
```

I want to finish this PR as adding just an image loader.

ssi02014: @HoseungJang 
I get it!

>I want to finish this PR as adding just an image loader.

Are you referring to the "useImageLoad" example you mentioned?

HoseungJang:
The conversation is going on for a long time.

Please make an issue first next time before you make a completely new feature like `useLazyImage`, to discuss about it.

It's helpful to determine not only if the new feature is appropriate but also how to develop it.

packages/react/react/src/hooks/useLazyImage.ts
HoseungJang: Yes, that's right.

ssi02014:
@HoseungJang 
That's right. I was too far ahead of myself, and I apologize for that.

I will work on it after discussing it after registering it in an issue when adding new features in the future.

PR #283, opened 2023-07-04
feat(@toss/react): Add root Props ImpressionArea

ssi02014:
## Overview
I worked on adding "root" props to the "Impression Area".
"root" props is one of the default options for "Intersection Observer". How about adding a corresponding "root" prop? 

I think it's better to cover a wider range of cases by providing a default option that can be utilized.

For example, The root option has issues with iframes. 

Of course, by default, the Viewport is applied as root (null), which I agree is appropriate in most cases. So, it would be great if you could review it.

Also, I don't think the code I've been working on will cause problems with older versions of the code.  🤗

## PR Checklist


ChanhyukPark-Tech (commented):
This is a good improvement.

What kind of situation did you actually need to give a root to?(in real programming)

Could you write a test code to verify this change?

ssi02014:
@ChanhyukPark-Tech 
In fact, root is an option that most of the time is not needed in real programming..!

However, I recently had an experience with `storybook` where rootMargin didn't work because I didn't specify root. 
when I specified root as document, "rootMargin" worked fine.

Take a look at the Intersection Observer option in the video below. 🙏


## root: null (Unusual behavior 🥲)
https://github.com/toss/slash/assets/64779472/d6775a99-c89c-4acf-94f8-496718885cfc

## root: document (Normal behavior 😄)
https://github.com/toss/slash/assets/64779472/8da1aea3-a6e1-4718-8ad3-c6c1985121ce

ssi02014:
@ChanhyukPark-Tech 
After my initial comment, I made a correction. 😂 If you're replying, please refresh!!

ChanhyukPark-Tech (approved):
Thank you for preparing a good example and reference and producing this pr. 

Someone might need this option

ssi02014:
@ChanhyukPark-Tech 
I'm glad this pull request got approved and merged.
Also, thanks for your review. 🙂

ChanhyukPark-Tech:
@ssi02014 
I learn a lot from you because you always create good PR. Thank you!😄

Issue #235, opened 2023-05-06
[Feature]: reverseKeyValue when object has multiple value.


ChanhyukPark-Tech:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: [@toss/utils reverseKeyValue](https://slash.page/libraries/common/utils/src/reverseKeyValue.i18n)

### **Overview**

below is always correct
```tsx
expect(reverseKeyValue({ jbee: 'eebj' })).toEqual({ eebj: 'jbee' });
```

but how about this object which is multiple same value

```tsx
reverseKeyValue({ jbee: 'eebj',jbee2:'eebj' }) // { "eebj": "jbee2"} 
```

It is not bug,, but people can make human error I think

### **Describe the solution you'd like**

Why don't we put the same values together in an array?

```tsx
reverseKeyValue({ jbee: 'eebj',jbee2:'eebj' }) // { "eebj": ["jbee","jbee2"]} 
```

### **Additional context**

In the case above, please suggest if it can be handled in another good way👍

wooki-kim:
안녕하세요 기여를 해보려고 구조를 파악중입니다.
질문이 있어요. 해당 파일의 유닛 테스트는 어떻게 해야할까요?
`yarn workspace @toss/utils test:unit` 명령어를 쓰니 아무런 반응이 없습니다. 혹시 제가 잘못 사용하고 있는지 궁금합니다.

ChanhyukPark-Tech:
Thank you for your interest.
I am also Korean, but since the rules of this repository are to communicate in English, I will follow the rules and answer them in English.

Did you install the appropriate package through `yarn install`? 
I run the command you showed me(`yarn workspace @toss/utils test:unit`), and the test works normally. (From the root path of the project)

https://github.com/toss/slash/assets/69495129/6f2bf8a7-ebd5-475e-936d-7f67504112f7

wooki-kim:
Thank you for your response and providing the information. Although the exact cause is unknown, it seems that the issue was resolved after upgrading Yarn.

Congratulations on submitting your first open-source PR! I have reviewed your PR at https://github.com/toss/slash/pull/250. I will provide feedback and suggestions for improvement promptly.

wooki-kim:
Thank you for your response. I appreciate it.
After upgrading yarn and reconfiguring it with yarn berry, the issue has been resolved.
I have submitted my first open source PR. You can find it here: https://github.com/toss/slash/pull/250
[...]

PR #284, opened 2023-07-05
feat(@toss/utils): Fix reverseKeyValue when object has multiple value.

wooki-kim:
## Overview
This PR aims to address the issue raised in https://github.com/toss/slash/issues/235

## PR Checklist

- [ ] Please verify if the proposed changes address the issue.
- [ ] Please ensure that the test cases and documentation are written correctly.
- [ ] Please verify if there are no performance issues.


packages/common/utils/src/reverseKeyValue.ts
```diff
@@ -1,4 +1,14 @@
 /** @tossdocs-ignore */
+
 export function reverseKeyValue<KeyType extends string, ValueType extends string>(obj: Record<KeyType, ValueType>) {
-  return Object.fromEntries(Object.entries(obj).map(([key, value]) => [value, key])) as Record<ValueType, KeyType>;
+  const reversedObj: Record<ValueType, KeyType | KeyType[]> = {} as Record<ValueType, KeyType | KeyType[]>;
```
minsoo-web, line 4:
```suggestion
  const reversedObj = {} as Record<ValueType, KeyType | KeyType[]>;
```

I think this type define is not required, cause `as` assertion is used.

packages/common/utils/src/reverseKeyValue.ts
```diff
@@ -1,4 +1,14 @@
...
+  for (const [key, value] of Object.entries(obj) as Array<[KeyType, ValueType]>) {
+    if (value in reversedObj) {
+      const existingValue = reversedObj[value];
+      reversedObj[value] = Array.isArray(existingValue) ? [...existingValue, key] : [existingValue, key];
+    } else {
+      reversedObj[value] = key;
+    }
+  }
+  return reversedObj;
```
minsoo-web, line 13: I think this API`s purpose not change object format, but revert key with value.

Since the object with the same key value is overwritten at the end, I think it is safe to say that the current return value is not a bug but an intended action.

image

So, I think this return value is not wanted by user.

wooki-kim: Thank you for your response. I sincerely appreciate it.
I consider this reply to be regarding the issue addressed in https://github.com/toss/slash/issues/235.
Unfortunately, I had to proceed with the task without knowledge of the initial intention and history of the method.
The fact that the issue from May is still open led me to believe that assistance was desired.
In the future, when making contributions to this open-source project, should I seek feedback on the desired type of PR in the corresponding issue before proceeding?
I am completely fine if this PR does not get accepted.
I believe that contributing to the well-structured monorepo of 'slash' would not only enhance my learning experience but also be beneficial to me.
If it is deemed unnecessary, would it be acceptable to close the PR?

ChanhyukPark-Tech: >Since the object with the same key value is overwritten at the end, I think it is safe to say that the current return value is not a bug but an intended action.

**I think it is good solution.**

If there is a duplicate key value, it seems to be the intention of object in `javascript`, seeing that it is overwritten with the last value used in order

PR #285, opened 2023-07-05
feat(@toss/utils): add isSubset function for array utils

kim-elijah-sol:
## Overview

Add utility function `isSubset`
This function is useful for checking if the first array is a subset of the second array.

## PR Checklist


packages/common/utils/src/array/isSubset.ts
```diff
@@ -0,0 +1,5 @@
+/** @tossdocs-ignore */
+
+export function isSubset(a: unknown[] = [], b: unknown[] = []): boolean {
+  return a.every(element => b.includes(element));
+}
```
minsoo-web, line 5: Could you tell me why you use default parameter?
this function has no JSDoc, and parameter's name is `a` and `b`
user has no idea what function it is, and must be check Docs.

image

> What can I do next ..? 🤔

I think it is more convenient to use parameter name such as `compared`, or `subSet` and `wholeSet` or `comparision` for user.

And Second suggest is deleting default parameter, make user to know who use typescript what is wrong, and check function's parameter or JSDoc

For example, my suggestion is below

image

1. delete default parameter
 user know it's function has two arguments

image

2. parameter's name is more readable

JSDoc is make more readable for user what this function it is.

image

minsoo-web (commented):
It is my pleasure for reviewing this PR 🧤
I wrote down things that I thought would be better.

packages/common/utils/src/array/isSubset.ts
```diff
@@ -0,0 +1,5 @@
+/** @tossdocs-ignore */
+
+export function isSubset(a: unknown[] = [], b: unknown[] = []): boolean {
+  return a.every(element => b.includes(element));
```
minsoo-web, line 4: Oh, I missed this

`includes` methods does not distinguish object type correctly

edge case is below

```ts
const foo = [[1,2,3], [4,5,6]];
const bar = [[1,2,3], [4,5,6], [7,8,9]];

isSubset(foo, bar); // Expected true, but false


const foo2 = [{a: 1}, {b: 1}];
const bar2 = [{a: 1}, {b: 1}, {c: 1}];

isSubset(foo2, bar2); // Expected true, but false

const foo3 = [new Set('1'), new Set('2')];
const bar3 = [new Set('1'), new Set('2'), new Set('3')];

isSubset(foo3, bar3); // Expected true, but false
```

If purpose of this function is compare of whole unknown type, you should make isEqual methods for unknown type such as lodash.isEqual

So, in my opinion, It is more convenient not unknown but `string | number`.  for simply find subset array using `includes` method

packages/common/utils/src/array/isSubset.ts
kim-elijah-sol: I appreciate your review!

It was my mistake to use default parameters, i will fix that.
and
Using JSDoc would make it easier for users to understand this function!

Thank you for the great idea 😃

packages/common/utils/src/array/isSubset.ts
kim-elijah-sol: For simple comparison, `string | number` seems more suitable than `unknown`.

Thank you for the great idea 😃

packages/common/utils/src/array/isSubset.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param wholeSet A set of comparison for `subSet`
+ *
+ * @returns boolean ( `subSet` is a subset of `wholeSet` )
+ *
+ * @example
+ * isSubset([], [1,2,3]); // true
+ * isSubset([1,2], [1,2,3]); // true
+ * isSubset([1,2,4], [1,2,3]); // false
+ */
+export function isSubset(subSet: Array<string | number>, wholeSet: Array<string | number>): boolean {
```
ChanhyukPark-Tech, line 18: Will this utility function be used frequently even if we limit the element type of the array to the `string and number` types? I think object type is a lot in real life😄

Can't you implement a utility function that can also determine the subset between objects?

### Example

```tsx
const foo = {
    a:1,
    b:2
}

const foo2 = {
    a:1,
    b:2
}

const foo3 = {
    a:1,
    b:3
}
console.log(isSubset([foo],[foo2,foo3]))
```

minsoo-web: I agree about this, but as I commented https://github.com/toss/slash/pull/285#discussion_r1253923197

> If purpose of this function is compare of whole unknown type, you should make isEqual methods for unknown type such as lodash.isEqual

If this function want to correspond to more than the `object` type, it is inevitable to create a function such as isEqual.

So, only `string | number` type arguments is not much more convenient, `isEqual` like function must be created first.

kim-elijah-sol: @ChanhyukPark-Tech
I appreciate your review!

in `@toss/react/useFunnel` use `fast-deep-equal` library for object comparison.

Similarly, it would be nice to use the `fast-deep-equal` library for object comparison for the `isSubset` function.

What do you think?

ChanhyukPark-Tech: @kim-elijah-sol sorry for late answer

I think `fast-deep-equal` is right solution.

is it good to accept only Array type ? how about accept Iterable type? is it over spec??]

this is my pseudocode

```tsx
import { equal } from 'fast-deep-equal';

function isSubset<T>(iterable1: Iterable<T>, iterable2: Iterable<T>): boolean {
  const array1 = Array.from(iterable1);
  const array2 = Array.from(iterable2);

  return array1.every(item => array2.some(otherItem => equal(item, otherItem)));
}

```

ChanhyukPark-Tech: I agree that receiving all `Iterable` types is over-spec.
I think it'll be okay to just get the array!

packages/common/utils/src/array/isSubset.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @returns boolean ( `subSet` is a subset of `wholeSet` )
+ *
+ * @example
+ * isSubset([], [1,2,3]); // true
+ * isSubset([1,2], [1,2,3]); // true
+ * isSubset([{a: 1}, {c: 3}], [{a: 1}, {b: 2}, {c: 3}]) // true
+ * isSubset([1,2,4], [1,2,3]); // false
+ */
+export function isSubset(subSet: unknown[], wholeSet: unknown[]): boolean {
```
okinawaa, line 20: I found subset is one word

```suggestion
export function isSubset(subset: unknown[], wholeSet: unknown[]): boolean {
```

packages/common/utils/src/array/isSubsest.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    // Given
+    const value1 = [] as string[];
+    const value2 = [1, 2, 3];
+
+    // When
+    const result = isSubset(value1, value2);
+
+    // Then
+    expect(result).toBe(true);
+  });
```
okinawaa, line 14:
```suggestion
  it('should return true if the first array is empty', () => {
    // Given
    const subset = [] as string[];
    const wholeSet = [1, 2, 3];

    // When
    const result = isSubset(subset, wholeSet);

    // Then
    expect(result).toBe(true);
  });
```

kim-elijah-sol: @okinawaa 
thank you for your review! 🙏

i understand your suggestion.
so i will fix to parameter name more exactly 🛠️

okinawaa (approved):
LGTM~!

raon0211:
Hello, thanks for your contribution!

We are really sorry to merge this pull request so late. Would you mind opening a new one after resolving the conflicts here? We would merge it right away after the conflicts has been merged.

Closing this pull request for this reason, but we will merge this as soon as you resolve the conflicts and open the pull request again :)

PR #157, opened 2022-11-24
feat(error-boundary): add ErrorBoundaryGroup

manudeli:
#### Referencing Issue
#144 

I Added ErrorBoundary.ResetKey, ResetKey, useResetKey, withResetKey to reset ErrorBoundary outside of itself easily.

### Problem
Resetting ErrorBoundary outside of itself is tiresome
```tsx
const Example = () => {
  const [key, setKey] = useState(0)
  const resetKey = () => setKey(prev => prev + 1)

  return <>
    <NestedResetter resetKey={resetKey} />
    <ErrorBoundary resetKeys={[key]}>
      <ThrowComponent>
    </ErrorBoundary>
    <NestedErrorBoundary resetKey={resetKey} /> {/* prop drilling is necessary */}
  </>
}
```

### Solution
if use useResetKey or ErrorBoundary.ResetKey, prop drilling is not necessary.
```tsx
const Example = withResetKey(() => (
  <>
    <NestedResetter />
    <ErrorBoundary.ResetKey>
      <ThrowComponent>
    </ErrorBoundary>
    <NestedErrorBoundary /> {/* prop drilling is not necessary */}
  </>
))
```

### Using withResetKey
```tsx
const Example = withResetKey(({ reset: resetAtOnce, resetKey }) => {
  const [ifYouWantToAddOtherDependency] = useState(false);

  return (
    <>
      <NestedResetter /> {/* Accessing ResetKeyContext is available by using useResetKey */}

      <button onClick={resetAtOnce}>reset at once outside of ErrorBoundary</button> {/* Resetting is easy */}

      <ErrorBoundary.ResetKey
        resetKeys={[ifYouWantToAddOtherDependency]} // Adding any other resetKey in resetKeys of ErrorBoundary.ResetKey is available
        renderFallback={({ reset }) => <button onClick={reset}>reset only self</button>}
      >
        <ThrowComponent />
      </ErrorBoundary.ResetKey>

      <ErrorBoundary
        resetKeys={[resetKey]}
        renderFallback={({ reset }) => <button onClick={reset}>reset only self</button>}
      >
        <ThrowComponent />
      </ErrorBoundary>
    </>
  );
});
```

### Nested Component using useResetKey
```tsx
const NestedResetter = () => {
  const { reset } = useResetKey();

  return <button onClick={reset}>reset ErrorBoundary outside of ErrorBoundary</button>;
};
```

### Using ResetKey.Provider, ResetKey.Consumer

```tsx
const Example = () => (
[...]

PR #287, opened 2023-07-09
fix(error-boundary): remove ErrorBoundaryGroup's not on purpose reset because of isMounted's side effect

manudeli:
related with https://github.com/suspensive/react/pull/92

## Remove ErrorBoundaryGroup's not on purpose reset because of isMounted's side effect

useIsMounted of @toss/react is implemented by useState, but we don't need to use useIsMounted triggering rerender component self when after mount.

### `useIsMounted` of @toss/react

```tsx
import { useEffect, useState } from 'react';

/** @tossdocs-ignore */
export function useIsMounted() {
  const [mounted, setMounted] = useState(false);
  useEffect(() => {
    setMounted(true);
  }, []);
  return mounted;
}

```

before this Pull Request, ErrorBoundaryGroup used useIsMounted to prevent reseting ErrorBoundaryGroup self on first rendering, and after mounting and checked whether parent ErrorBoundaryGroup.resetKey is changed to reset self.

but after this Pull Request is merging, we don't need to make unnecessary rerendering due to useIsMounted.

### In addition, we can prevent unintended ErrorBoundaryGroup resetting by useIsMounted.

[When first I add this ErrorBoundaryGroup in @toss/error-boundary](https://github.com/toss/slash/pull/157), I couldn't consider useIsMounted's rerendering side effect

#### AS-IS
ErrorBoundaryGroup will trigger reset on render because of useIsMounted

image

#### TO-BE
ErrorBoundaryGroup won't trigger reset on render because of removed useIsMounted

image

Check more detail and procedure to add this Pull Request in https://github.com/suspensive/react/pull/92 please

## PR Checklist


packages/react/error-boundary/src/ErrorBoundary.tsx
```diff
@@ -51,7 +51,6 @@ const initialState: ErrorBoundaryState = {
 
 class BaseErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
   state = initialState;
-  updatedWithError = false;
```
manudeli, line 54: I thought this updatedWithError is unnecessary field, there is no updatedWithError in https://github.com/bvaughn/react-error-boundary/blob/master/src/ErrorBoundary.ts too.

If this is matter, let me know why meaningful please.

packages/react/error-boundary/src/ErrorBoundary.tsx
```diff
@@ -67,31 +66,17 @@ class BaseErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState
     onError?.(error, info);
   }
 
-  resetState() {
-    this.updatedWithError = false;
-    this.setState(initialState);
-  }
-
```
manudeli, line 74: I remove this method. if no updatedWithError field, resetState is not anymore meaningful too.
I thought `this.setState(initialState)` is better to express what this is.

packages/react/error-boundary/src/hooks/useIsChanged.ts
```diff
@@ -0,0 +1,4 @@
+/** @tossdocs-ignore */
+import { usePrevious } from '@toss/react';
+
+export const useIsChanged = (value: unknown) => usePrevious(value) !== value;
```
manudeli, line 4: I made custom hook, `useIsChanged` to check whether value is changed with previous rendering value.

this hook will check parent ErrorBoundaryGroup.resetKey is changed
If parentErrorBoundaryGroup.resetKey is changed, ErrorBoundaryGroup self will change resetkey to propagate to child ErrorBoundaryGroup recursively.

okinawaa (approved):
LGTM👍🚀

PR #290, opened 2023-07-14
fix(@toss/utils): refactor isMobile.ts using getOSByUserAgent.ts

sa02045:
## Overview
Refactor isMobile.ts using getOSByUserAgent.ts

pros
- Unify the level of abstraction inside a function.
- Create dependencies between functions that are related to each other

### When supporting different mobile OS

When supporting different mobile OS, not using getOSByUserAgent.ts increases duplication and complexity.

```js
// getOsByUserAgent.ts
export function getOSByUserAgent() {
  // ...
  // ...
  const isBlackBerry = window.navigator.userAgent.match(/BlackBerry/i) !== null; <- duplication

  if (isBlackBerry) {
    return 'blackberry';
  }

  return 'web';
}


// isMobileWeb.ts
import { isServer } from './isServer';

export function isMobileWeb() {
  if (isServer()) {
    return false;
  }

  if (
    navigator.userAgent.match(/ipad|iphone/i) !== null ||
    navigator.userAgent.match(/Android/i) !== null ||
    navigator.userAgent.match(/BlackBerry/i) !== null.  <- duplication
  ) {
    return true;
  }
  return false;
}

```

using getOSByUserAgent

```js
/** @tossdocs-ignore */
import { getOSByUserAgent } from '.';
export function isMobileWeb() {
  const userAgent = getOSByUserAgent();

  if (userAgent === 'ios' || userAgent === 'android' || userAgent === 'blackberry') {
    return true;
  }

  return false;
}

```
## PR Checklist

### DIFF
--- a/packages/common/utils/src/device/isMobileWeb.ts
+++ b/packages/common/utils/src/device/isMobileWeb.ts
@@ -1,12 +1,10 @@
 /** @tossdocs-ignore */
-import { isServer } from './isServer';
+import { getOSByUserAgent } from './getOSByUserAgent';
 
 export function isMobileWeb() {
-  if (isServer()) {
-    return false;
-  }
+  const userAgent = getOSByUserAgent();
 
-  if (navigator.userAgent.match(/ipad|iphone/i) !== null || navigator.userAgent.match(/Android/i) !== null) {
+  if (userAgent === 'ios' || userAgent === 'android') {
     return true;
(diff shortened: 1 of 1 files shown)


ChanhyukPark-Tech (approved):
Thank you for the detailed explanation.
It's nice to be able to use existing defined functions.

PR #291, opened 2023-07-17
feat(@toss/react): Add useImageLoad Hook

ssi02014:
## Overview

https://github.com/toss/slash/pull/276  
I worked on the `useImageLoad` hook mentioned in that pull request. 

The `useImageLoad` hook allows you to check whether an image resource has been successfully loaded with the return value 'isLoaded' for an image tag provided with `ref`.

This is a pretty small module and can be utilized in conjunction with a number of other modules by the user (e.g. `ImpressionArea`).

However, I was unable to write test code for that hook...🥲
It's very difficult to write test code for the image loading completion.

## PR Checklist


ssi02014:
@HoseungJang 
Speaking of code, the "useImageLoad" example in the [#276](https://github.com/toss/slash/pull/276) pull request has the inconvenience of having to pass the image source to both the "hook" and "img" tags.

There's also the problem that you might be passing different src's to both. I improved this by using "ref".

By only passing "src" and "ref" to the "img" tag, we can get the isLoaded value we want.👍

packages/react/react/src/hooks/useImageLoad.ts
```diff
@@ -0,0 +1,46 @@
+import { noop } from '@toss/utils';
+import { useCallback, useState } from 'react';
+import { usePreservedCallback } from './usePreservedCallback';
+
+interface Options {
+  onLoadComplete?: () => void;
+  onLoadStart?: () => void;
```
ChanhyukPark-Tech, line 7: It's very minor, but it would be nice to arrange it according to the order in which it is used!

```suggestion
  onLoadStart?: () => void;
  onLoadComplete?: () => void;
```

ChanhyukPark-Tech (approved):
I think it's a good hook.
@HoseungJang  who helped us a lot, please give final review!

thank you

ssi02014:
@ChanhyukPark-Tech 
Thank you for the review! 🙏

PR #292, opened 2023-07-19
fix(@toss/react): literal type inference

ojj1123:
## Overview

#220 

**I wrote a PR by modifying the type for `defaultValue` so that `generic` is not forced.**

TypeScript will infer **"a concrete type"** if there is a `generic type constraint` for a `primitive type`.
`Type constraint` is applied to `useStorageState` so that only values that can be serialized as `JSON` can be received as `defaultValue`.

However, if `defaultValue` was set as a `primitive type`, the `state` was inferred to be a `literal type`, and `generic` had to be created in the following situations.
```ts
// No Generic
  const [state, setState, refresh] = useStorageState('key', {
    defaultValue: 0,
  });

  useEffect(() => {
    setState(x => x + 1);
    // ^ type error
  }, []);

// Using Generic
  const [state, setState, refresh] = useStorageState<number>('key', {
    defaultValue: 0,
  });

  useEffect(() => {
    setState(x => x + 1); // No Error
  }, []);
```

### How To Fix
**The `Serializable` type was inferred by dividing it into a `primitive type` and an `object type`.**
`Primitive types` must be inferred as `primitive types`, **not `literal types`** (See `ToPrimitive<T>`), and `object types` are shallowly inferred, so it's okay to infer `T` when `true` for a conditional type (See `ToObject<T>`).
```ts
type ToPrimitive<T> = T extends string ? string : T extends number ? number : T extends boolean ? boolean : never;
type ToObject<T> = T extends unknown[] | Record<string, unknown> ? T : never;
export type Serializable<T> = T extends string | number | boolean ? ToPrimitive<T> : ToObject<T>;
```

p.s. Tried to solve this PR #223 as well, but `state` was still inferred as a `literal type`.

### AS-IS
```ts
// Need generic
  const [state, setState, refresh] = useStorageState('@service/some-resource', {
    defaultValue: 0,
  });

  state; // 0
```
### TO-BE
```ts
// No using generic
  const [state, setState, refresh] = useStorageState('@service/some-resource', {
    defaultValue: 0,
  });

  state; // number
```
## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useStorageState.spec.ts
+++ b/packages/react/react/src/hooks/useStorageState.spec.ts
@@ -141,7 +141,7 @@ function createMockStorage() {
   };
 }
 
-function createFixture<T extends Serializable>({ defaultValue }: { defaultValue?: T } = {}) {
+function createFixture<T>({ defaultValue }: { defaultValue?: Serializable<T> } = {}) {
   const key = '@@test-key';
   const storage = createMockStorage();
   const render = () => renderHook(() => useStorageState<T>(key, { storage, defaultValue }));
--- a/packages/react/react/src/hooks/useStorageState.ts
+++ b/packages/react/react/src/hooks/useStorageState.ts
@@ -2,35 +2,39 @@
 import { safeLocalStorage, Storage } from '@toss/storage';
 import { SetStateAction, useCallback, useState } from 'react';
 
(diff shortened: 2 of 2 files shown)


raon0211 (approved):
Thanks, @ojj1123 !

PR #295, opened 2023-07-22
feat(@toss/use-query-param): Implement required option in useQueryParam hook.

okinawaa:
# Merge @tossteam/use-query-param

## Overview
I added the `required` option to the `useQueryParam` hook. When using this option, if the desired query parameter does not exist, it will throw an error. If the query parameter exists, it will be treated as `NonNullable`.

## PR Checklist


packages/react/use-query-param/src/hooks/useQueryParam.spec.ts
```diff
@@ -1,16 +1,16 @@
...
+import { useQueryParam } from './useQueryParam';
 
 jest.mock('next/router');
 
 const mockRouter = mocked(Router);
 const mockUseRouter = mocked(useRouter as () => Partial<NextRouter>);
 
 describe('useQueryParam', () => {
-  it('첫 번째 파라미터와 일치하는 이름의 쿼리 파라미터를 읽을 수 있다.', () => {
+  it('can read the query parameter with a name matching the first parameter.', () => {
```
ChanhyukPark-Tech, line 13: While implementing the function, I also changed the description of the test code from Korean to English.

raon0211 (approved):

PR #296, opened 2023-07-22
fix(@toss/react, @toss/use-loading): implement `useIsMountedRef` and modified `useLoading` to avoid state updates when the component is unmounted

okinawaa:
## Overview

# Merge @tossteam/use-loading

# @toss/react

Implemented `useIsMountedRef`

### Motivation([reference](https://github.com/helderberto/use-is-mounted-ref))
- Avoid memory leaks setting states when component are unmounted;
- Control when component already mounted;
- Common error when setting state to unmounted component:

# @toss/use-loading

## Preventing State Updates After Unmounting in useLoading Hook

I have made changes to the `useLoading` hook to address the issue where the state gets updated after the component calling useLoading is unmounted. This problem typically occurs when useLoading receives a promise that might lead to the unmounting of the component. When the promise resolves after the component is already unmounted, it triggers a "Can't perform a React state update on an unmounted component" warning.

The revised version of the useLoading hook ensures a more reliable and stable behavior when dealing with asynchronous tasks and component unmounting.

## PR Checklist


packages/react/use-loading/src/useLoading.ts
```diff
@@ -1,16 +1,37 @@
...
+
+  useEffect(() => {
+    ref.isMounted = true;
+    return () => {
+      ref.isMounted = false;
+    };
+  }, [ref]);
+
+  return ref;
+}
```
minsoo-web, line 37: Why don't you add this hook to `@toss/react` ?!

It is pretty cool hook to user who manage state regardless rendering lifecycle

ChanhyukPark-Tech: thx I think it's a `hook` that can be used to prevent unwanted state updates in many situations

raon0211 (approved):

PR #299, opened 2023-07-23
test(@toss/react): Add useIsMounted Test Code

ssi02014:
## Overview
I wrote some test code that is very simple, but needs to return true when the useIsMounted Hook is mounted. 🙏

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useIsMounted.spec.tsx
+++ b/packages/react/react/src/hooks/useIsMounted.spec.tsx
@@ -1,0 +1,10 @@
+import { renderHook } from '@testing-library/react';
+import { useIsMounted } from './useIsMounted';
+
+describe('useIsMounted', () => {
+  it('should return "true" when mounted', () => {
+    const { result } = renderHook(() => useIsMounted());
+
+    expect(result.current).toBe(true);
+  });
+});


ChanhyukPark-Tech (changes requested):
I think we need to test **two things** to verify the `useMountedHook`.

- return false initial value
- return true value after mounting.

I think you only tested the `second case`, so could you test the first case as well?

ssi02014:
@ChanhyukPark-Tech 
Unfortunately, I tried several times to test the default value of "false" for "mounted" and failed. 🥲

Does anyone have any good ideas for testing initialization?

I didn't find it first, and consulted libraries like "usehooks-ts".
[usehooks-ts/useIsMounted](https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useIsMounted/useIsMounted.ts)
[usehooks-ts/useIsMounted.test](https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useIsMounted/useIsMounted.test.ts)

That hook also doesn't test for default values, it only tests for "mount" and "unmount" situations.

However, currently toss/react's useIsMounted is managed by useState, so testing for the unmounted case is also limited...😭 (I think it's too burdensome to change the useIsMounted. )

I think it's a shame that there is no test code for default values. 

However, I think it's better to have test code for the "mount" situation than no test code at all.

ChanhyukPark-Tech (approved):
Since `@testing-library/react` is a library focused on rendering and interacting with actual react components, it is not supported to directly test the condition before the components are mounted.

Thank you for writing a good test code

ssi02014:
@ChanhyukPark-Tech 
Yes, that's right! Thank you for the review 🎉

PR #300, opened 2023-07-23
test(@toss/react): Add usePreservedReference Test Code

ssi02014:
## Overview

I wrote test code for a usePreservedReference that didn't exist before. 🙏

```tsx
it('should return the initial value', () => {
  const initialValue = { name: 'toss' };

  const { result } = renderHook(({ value }) => usePreservedReference(value), {
    initialProps: { value: initialValue },
  });

  expect(result.current).toBe(initialValue); // I used "toBe" because it's the same reference.
});
```

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/usePreservedReference.spec.tsx
+++ b/packages/react/react/src/hooks/usePreservedReference.spec.tsx
@@ -1,0 +1,62 @@
+import { renderHook } from '@testing-library/react';
+import { usePreservedReference } from './usePreservedReference';
+
+describe('usePreservedReference', () => {
+  it('should return the initial value', () => {
+    const initialValue = { name: 'toss' };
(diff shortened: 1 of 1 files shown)


ChanhyukPark-Tech:
How about see #250 ?
Is it different task?

ssi02014:
@ChanhyukPark-Tech 
Oh, a similar pull request already existed. 😂
This Pull request has the same purpose as #250.
However, there is a difference in the way the test code is written.

ChanhyukPark-Tech:
> However, there is a difference in the way the test code is written.

I hope we can discuss it at #250 and take a better way

ssi02014:
@ChanhyukPark-Tech 
Yes! I agree.

PR #301, opened 2023-07-24
fix(error-boundary): Merge @tossteam/error-boundary

HoseungJang:
## PR Checklist

### DIFF
--- a/packages/react/error-boundary/src/ErrorBoundary.tsx
+++ b/packages/react/error-boundary/src/ErrorBoundary.tsx
@@ -19,7 +19,7 @@ import { ComponentPropsWithoutChildren } from './types/index';
 
 type RenderFallbackProps<ErrorType extends Error = Error> = {
   error: ErrorType;
-  reset: (...args: unknown[]) => void;
+  reset: () => void;
 };
 
 type RenderFallbackType = <ErrorType extends Error>(props: RenderFallbackProps<ErrorType>) => ReactNode;
--- a/packages/react/error-boundary/src/ErrorBoundary.tsx
+++ b/packages/react/error-boundary/src/ErrorBoundary.tsx
@@ -15,41 +15,39 @@ import {
   useState,
 } from 'react';
 import { ErrorBoundaryGroupContext } from './ErrorBoundaryGroup';
-import { ComponentPropsWithoutChildren } from './types';
+import { ComponentPropsWithoutChildren } from './types/index';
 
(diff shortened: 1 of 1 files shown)


ChanhyukPark-Tech (approved):
good👍

packages/react/error-boundary/src/ErrorBoundary.tsx
```diff
@@ -25,31 +25,29 @@ type RenderFallbackProps<ErrorType extends Error = Error> = {
...
   error: ErrorType | null;
 }
 
-const initialState: ErrorBoundaryState = {
+const initialState: State = {
   error: null,
 };
 
-class BaseErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
+class BaseErrorBoundary extends Component<PropsWithRef<PropsWithChildren<Props>>, State> {
```
manudeli, line 50: How about maintain this lines as previous implemantation?

1. I think ErrorBoundaryProps is better name than Props 
2. ErrorBoundaryProps need to contain children

### [Discussion with react-error-boundary maintainer](https://github.com/bvaughn/react-error-boundary/pull/152)
It's not mandatory following react-error-boundary, but it's a bit sad to make this change only merging and not intuitive change.

packages/react/error-boundary/src/ErrorBoundary.tsx
```diff
@@ -77,7 +75,7 @@ class BaseErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState
     this.resetState();
   };
 
-  componentDidUpdate(prevProps: ErrorBoundaryProps) {
+  componentDidUpdate(prevProps: Props) {
```
manudeli, line 78: I want to leave comment with this link

https://github.com/bvaughn/react-error-boundary/pull/152#discussion_r1272742523

manudeli (commented):
I knew this Pull Request is for merging with @tossteam/libraries, but it's a little sad if this change is merged

packages/react/error-boundary/src/ErrorBoundary.tsx
HoseungJang: 1. I think it's not important because the type isn't exported.
2. So it's wrapped by `PropsWithChildren`.

packages/react/error-boundary/src/ErrorBoundary.tsx
HoseungJang: I know what you consider but this is not necessary in our case.

PR #318, opened 2023-08-03
Feat : useDeviceInfo

ChanghyeonYoon:
## PR Checklist


ChanghyeonYoon:
ISSUE #317

ChanghyeonYoon:
Jest 환경에서 window.navigator.userAgent를 확인할 수 없어 test코드는 추가하지 않았습니다.

packages/react/react/src/hooks/useDeviceInfo.ts
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+/** @tossdocs-ignore */
+export const useDeviceInfo = ()=> {
```
minsoo-web, line 4: I think this hook pretty useful to react developer.
But It is already in slash library utils

Plz check links below

[isClient](https://slash.page/libraries/common/utils/src/device/isClient.i18n)
[isMacOS](https://slash.page/libraries/common/utils/src/device/isMacOS.i18n)
[isMobileWeb](https://slash.page/libraries/common/utils/src/device/isMobileWeb.i18n)
[isServer](https://slash.page/libraries/common/utils/src/device/isServer.i18n)
[getOSByUserAgent](https://slash.page/libraries/common/utils/src/device/getOSByUserAgent.i18n)

So, It seems a little unclear what are the advantages of using this hook, rather than directly using the toss/utils function.
There seems to be a little difference between whether to use a util function separately to get the information you need and whether to call a hook that drops all the information at once.

Alternatively, if it is necessary to create a flexible hook for changing the state of react, it is likely that taking and using the utility function of the toss will not divide the management points

PR #319, opened 2023-08-03
fix(@toss/utils): Merge with internal `@tossteam/utils` library

okinawaa:
## Overview

It works differently

I got this image error

image

The two TypeScript type declarations perform fundamentally similar roles, but they result in different types. Here are the differences:

```tsx
export declare type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`;
```

This declaration converts all property keys of type T into string types. It forces the property keys into string types using Template Literal Types, which is a feature introduced in TypeScript 4.1. Consequently, this type becomes a union of string literals.

```tsx
export declare type ObjectKeys<T extends Record<PropertyKey, unknown>> = Exclude<keyof T, symbol>;
```
This declaration returns all property keys of type T but as keys themselves, not as strings. This type returns a union of all property keys of T excluding those of type symbol. Consequently, this type becomes a union of property keys.

Therefore, the primary difference is that the first declaration returns a union of string literals while the second one returns a union of property keys. Due to these differences, each declaration can be used for slightly different purposes.

## PR Checklist

### DIFF
--- a/packages/common/utils/src/object/object-keys.ts
+++ b/packages/common/utils/src/object/object-keys.ts
@@ -1,5 +1,5 @@
 /** @tossdocs-ignore */
-export type ObjectKeys<T extends Record<PropertyKey, unknown>> = Exclude<keyof T, symbol>;
+export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`;
 
 export function objectKeys<Type extends Record<PropertyKey, unknown>>(obj: Type): Array<ObjectKeys<Type>> {
   return Object.keys(obj) as Array<ObjectKeys<Type>>;


minsoo-web:
Can you give some example to reproduce your situation like screenshots?
I understand what you explained, but want to check reproducing...

okinawaa:
@minsoo-web 

you can simply copy and paste below code to [typescript playgrond](https://www.typescriptlang.org/play)

```tsx

// export type ObjectKeys<T extends Record<PropertyKey, unknown>> = Exclude<keyof T, symbol>; // it got error
export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`; // it have no error

function objectEntries<Type extends Record<PropertyKey, unknown>>(
  obj: Type
): Array<[ObjectKeys<Type>, Type[ObjectKeys<Type>]]> {
  return Object.entries(obj) as Array<[ObjectKeys<Type>, Type[ObjectKeys<Type>]]>;
}



function filterObjectValues<ObjectType extends Record<string, unknown>>(
  values: ObjectType,
  predicate: (value: ObjectType[keyof ObjectType]) => boolean
) {
  const result: Record<string, unknown> = {};

  objectEntries(values).forEach(([key, value]) => {
    if (predicate(value)) {
      result[key] = value;
    }
  });

  return result;
}

```

minsoo-web:
Thank you for kind example.

I think this is not because ObjectKeys type but your example case function's result type is Record<string, unknown>

```ts
function filterObjectValues<ObjectType extends Record<PropertyKey, unknown>>(
  values: ObjectType,
  predicate: (value: ObjectType[keyof ObjectType]) => boolean
) {
  const result: Record<PropertyKey, unknown> = {} // not Record<string, unkown>;

  objectEntries(values).forEach(([key, value]) => {
    if (predicate(value)) {
      result[key] = value;
    }
  });

  return result;
}

// use case

const foo = {
  a: '1',
  1: '2'
}

const filteredFoo =  filterObjectValues(foo, (value) => value === '1');
console.log(filteredFoo); // {a: '1'}
```

Here is my change suggestion.
there is no type error

okinawaa:
https://github.com/toss/slash/pull/61

I know that I can solve the `type error` by correcting the example. like you suggested

I think it's better to cover it with ``. We can type more flexibly.

Currently, there are many services that use the `objectEntries` function that was before [PR 61](https://github.com/toss/slash/pull/61), but I think it can make a big difference if there is a type error after updating new version. 

Because wrap with `` can provide a higher level of compatibility

In future When considering the `BREAKING CHANGE` later on, we'd better get rid of the ``.

PR #320, opened 2023-08-04
feat(@toss/react): add gap props to flex

SEOKKAMONI:
## Overview

I added a gap prop to flex.tsx on @toss/emotion-utills
People usually use flex and gap together. I think it could be a better code if Flex.tsx manages this together

By adding the 'gap' prop, we eliminate an unnecessary one line.
### AS-IS
```tsx
${flex({ align: 'center' })};
gap: 16px;
```

### TO-BE
```tsx
${flex({ align: 'center', gap: "16px" })};
```

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/flex.tsx
+++ b/packages/react/emotion-utils/src/flex.tsx
@@ -7,27 +7,31 @@ export interface FlexOptions {
   align?: CSSProperties['alignItems'];
   justify?: CSSProperties['justifyContent'];
   direction?: CSSProperties['flexDirection'];
+  gap?: CSSProperties['gap'];
 }
 
 export function flex(options: FlexOptions): SerializedStyles;
 export function flex(
   align: CSSProperties['alignItems'],
   justify?: CSSProperties['justifyContent'],
-  direction?: CSSProperties['flexDirection']
+  direction?: CSSProperties['flexDirection'],
+  gap?: CSSProperties['gap']
 ): SerializedStyles;
 export function flex(
   alignOrFlexOptions: FlexOptions | CSSProperties['alignItems'],
   justify = 'flex-start',
-  direction = 'row'
+  direction = 'row',
+  gap = 0
 ) {
   if (typeof alignOrFlexOptions === 'object') {
-    const { align = 'stretch', direction = 'row', justify = 'flex-start' } = alignOrFlexOptions;
+    const { align = 'stretch', direction = 'row', justify = 'flex-start', gap = 0 } = alignOrFlexOptions;
 
     return css`
       align-items: ${align};
       display: flex;
       flex-direction: ${direction};
       justify-content: ${justify};
+      gap: ${gap};
     `;
   }
 
@@ -36,6 +40,7 @@ export function flex(
     display: flex;
     flex-direction: ${direction};
     justify-content: ${justify};
+    gap: ${gap};
   `;
 }
 
@@ -54,10 +59,11 @@ const createFlexComponent = (flexOptions?: FlexOptions): FlexComponentType =>
       direction = flexOptions?.direction ?? 'row',
       justify = flexOptions?.justify ?? 'flex-start',
       align = flexOptions?.align ?? 'stretch',
+      gap = flexOptions?.gap ?? 0,
       ...rest
     } = props;
     const Component = as;
-    return <Component ref={ref} css={flex({ direction, align, justify })} {...rest} />;
+    return <Component ref={ref} css={flex({ direction, align, justify, gap })} {...rest} />;
   });
 
 type FlexType = FlexComponentType & {


packages/react/emotion-utils/src/flex.tsx
```diff
@@ -7,27 +7,31 @@ export interface FlexOptions {
...
-  direction?: CSSProperties['flexDirection']
+  direction?: CSSProperties['flexDirection'],
+  gap?: CSSProperties['gap']
 ): SerializedStyles;
 export function flex(
   alignOrFlexOptions: FlexOptions | CSSProperties['alignItems'],
   justify = 'flex-start',
-  direction = 'row'
+  direction = 'row',
+  gap = 0
```
okinawaa, line 24: `CSSProperties['gap']` can be  `Property.Gap<string | number> | undefined`

Is it okay to fix the default to zero?

packages/react/emotion-utils/src/flex.tsx
```diff
@@ -7,27 +7,31 @@ export interface FlexOptions {
...
   if (typeof alignOrFlexOptions === 'object') {
-    const { align = 'stretch', direction = 'row', justify = 'flex-start' } = alignOrFlexOptions;
+    const { align = 'stretch', direction = 'row', justify = 'flex-start', gap = 0 } = alignOrFlexOptions;
 
     return css`
       align-items: ${align};
       display: flex;
       flex-direction: ${direction};
       justify-content: ${justify};
+      gap: ${gap};
```
okinawaa, line 34: CSSProperties['gap'] can be Property.Gap<string | number> | undefined

It does not have to suffix `px` ?

What if the value already includes px?

okinawaa (commented):
Thank you for your contribution!

If you change the interface of an existing function or add a new function, **I want to discuss whether it is a necessary function first through [ISSUE](https://github.com/toss/slash/issues)!**

packages/react/emotion-utils/src/flex.tsx
SEOKKAMONI: The default of gap is that it doesn't need gap, so I don't think it'll be a problem even if it's 0

packages/react/emotion-utils/src/flex.tsx
SEOKKAMONI: I missed it. I think I need to think more about this

packages/react/emotion-utils/src/flex.tsx
minsoo-web: I think this could be type error for overriding

image

packages/react/emotion-utils/src/flex.tsx
minsoo-web: In my Opinion what about this?

```suggestion
      ${gap && `gap: ${gap}`}
```

There is no need to set default value, and user can be modify it by setting typed also
gap value is only setted by user's parameter

minsoo-web (commented):
And what about added gap example to docs (ko, en) and test case for gap

okinawaa:
How about using Stack? When hou have to use gap option

packages/react/emotion-utils/src/flex.tsx
SEOKKAMONI: Sounds like a very good code Thank you, minsu !!

SEOKKAMONI:
> How about using Stack? When hou have to use gap option

oh...  I'm surprised there was such a thing However, the React Stack component does not seem to support various spacing For example, units like % 

Here is the reference I consulted. If what I stated is not accurate, please provide comments.
> https://mui.com/material-ui/react-stack

packages/react/emotion-utils/src/flex.tsx
minsoo-web: I don't understand correctly, but I can check '12px' is apply in my code

I guess you worry about this situation

image

emotion's css function strip string value.
It is same case about

`justfy: 'center'` is apply `justify-contents: center` not `justify-contents: 'center'`

SEOKKAMONI: oh very cool !

minsoo-web:
> > How about using Stack? When hou have to use gap option
> 
> oh... I'm surprised there was such a thing However, the React Stack component does not seem to support various spacing For example, units like %
> 
> Here is the reference I consulted. If what I stated is not accurate, please provide comments.
> 
> > https://mui.com/material-ui/react-stack

I think 92914 said about this https://slash.page/libraries/react/emotion-utils/src/stack.i18n/

SEOKKAMONI:
> 

Ok I will try

SEOKKAMONI:
> 

Ok I will try

SEOKKAMONI:
> And what about added gap example to docs (ko, en) and test case for gap

Ok I will try

okinawaa:
[gap's support has been relatively recently started](https://caniuse.com/flexbox-gap) so it cannot support various users.

So the [Stack](https://slash.page/libraries/react/emotion-utils/src/stack.i18n/) component also adjusted the spacing of the elements in a way other than gap (use margin),

I think we can use Stack if we want to give a gap between elements, and whether we should provide this as a function rather than a component is a different matter.

SEOKKAMONI:
> cannot support various users.

I agree with you. gap is relatively recent and cannot support various users. but I think there are many reasons to make Gap a function I think it can be used when there are abstractions or various styles of elements that Stack components cannot do.
So please consider adding features later, not immediately

thank you for @okinawaa  @minsoo-web !!

PR #321, opened 2023-08-05
test(@toss/emotion-utils): Add FullHeight Component Test Code

ssi02014:
## Overview
I worked test code for the "FullHeight".
Please review this pull request to see if it's appropriate. 🙏

## References

- [Set Size of Window in Jest](https://stackoverflow.com/questions/60396600/set-size-of-window-in-jest-and-jest-dom-and-jsdom)

```ts
Object.defineProperty(window, 'innerWidth', { writable: true, configurable: true, value: 300 });
```

- [jsdom default Value (innerWidth: 1024, innerHeight: 768)](https://github.com/jsdom/jsdom/blob/0cba358253fd5530af0685ac48c2535992464d06/lib/jsdom/browser/Window.js#L587-L588)

스크린샷 2023-07-30 오전 2 54 09

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/FullHeight.spec.tsx
+++ b/packages/react/emotion-utils/src/FullHeight.spec.tsx
@@ -1,0 +1,31 @@
+import { render, screen } from '@testing-library/react';
+import { FullHeight } from './FullHeight';
+
+const setInnerHeight = (height: number) => {
+  Object.defineProperty(window, 'innerHeight', { writable: true, configurable: true, value: height });
+};
+
+describe('FullHeight', () => {
+  it('should have height 768px', () => {
+    render(<FullHeight>FullHeight</FullHeight>);
+
+    const fullHeightBox = screen.getByText('FullHeight');
+
+    expect(window.innerHeight).toBe(768);
+
+    expect(fullHeightBox).toBeInTheDocument();
+    expect(fullHeightBox).toHaveStyle('height: 768px');
+  });
+
+  it('should have height 1024px', () => {
+    setInnerHeight(1024);
+    render(<FullHeight>FullHeight</FullHeight>);
+
+    const fullHeightBox = screen.getByText('FullHeight');
+
+    expect(window.innerHeight).toBe(1024);
+
+    expect(fullHeightBox).toBeInTheDocument();
+    expect(fullHeightBox).toHaveStyle('height: 1024px');
+  });
+});
--- a/packages/react/emotion-utils/src/FullHeight.spec.tsx
+++ b/packages/react/emotion-utils/src/FullHeight.spec.tsx
@@ -8,7 +8,7 @@ const setInnerHeight = (height: number) => {
 const CHILDREN_TEXT = 'FullHeight';
 
 describe('FullHeight', () => {
-  it('should have height 768px', () => {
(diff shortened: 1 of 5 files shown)


packages/react/emotion-utils/src/FullHeight.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const setInnerHeight = (height: number) => {
+  Object.defineProperty(window, 'innerHeight', { writable: true, configurable: true, value: height });
+};
+
+describe('FullHeight', () => {
+  it('should have height 768px', () => {
+    render(<FullHeight>FullHeight</FullHeight>);
+
+    const fullHeightBox = screen.getByText('FullHeight');
```
okinawaa, line 12: how about seperate FullHeight text? 
It's not directly related to the test, but it's a means of testing.
and it can sustain SSOT

```tsx

const CHILDREN_TEXT = 'FullHeight'; // I'd like to change the variable name to something else


describe('FullHeight', () => {
  it('should have height 768px', () => {
    render(<FullHeight>{CHILDREN_TEXT}</FullHeight>);

    const fullHeightBox = screen.getByText(CHILDREN_TEXT);
```

okinawaa (changes requested):
In the context of testing, the current `768` and `1024` belong to the same test group. So there is no need to test twice. 769,780,781,782 .. Same reason not to test etc.

packages/react/emotion-utils/src/FullHeight.spec.tsx
ssi02014: @okinawaa 
Thank you for your feedback.
Your comments resonate with us. I will fix the code.

ssi02014:
@okinawaa 
>Because bug likes boundaries
Is there no need to test situations where height is 0 or negative?

The height of `FullHeight` depends on the internal `height` state, so the default value of height is `0`, so the case of 0 makes sense to consider. 

But is there a case where "window.innerHeight" could be `0` or `negative`? 

Actually, innerHeight is the `inner height value of the browser window`, which can get close to zero, but it's not likely to be 0 or negative.

So I don't think we need to deal with it. However, I'm open to adding a case of 0 if the majority thinks it's necessary.

(However, it would probably be code in much the same way as the 2nd test case).

packages/react/emotion-utils/src/FullHeight.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    const fullHeightBox = screen.getByText(CHILDREN_TEXT);
+
+    expect(window.innerHeight).toBe(768);
+
+    expect(fullHeightBox).toBeInTheDocument();
+    expect(fullHeightBox).toHaveStyle('height: 768px');
+  });
+
+  it('should have height 1024px', () => {
```
okinawaa, line 22: Could you write two different test descriptions differently😄

One is to follow the default height, and the other is to follow the height that you set, 
so what test codes want to verify is different

ssi02014: @okinawaa 
Thank you🙏 I fixed the code again.

okinawaa (approved):
cool thanks !

ssi02014:
@okinawaa Thanks 🤗🎉

PR #325, opened 2023-08-08
fix(@toss/react): Add dependency array to useTimeout and rename timeoutId

ssi02014:
## Overview
I added a dependency array to `useTimeout` and renamed it from `id` to `timeoutId` to give it a clearer meaning.

I also modified the example code in the `useTimeout` documentation.

After modifying the code, all tests passed! 🙏

https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value

## PR Checklist


raon0211 (approved):
Thanks!

PR #326, opened 2023-08-08
fix(@toss/date): unify rules

SEOKKAMONI:
## Overview

The rules were unified

**question: Is there a reason why only certain parts are arrow functions?**

## PR Checklist

### DIFF
--- a/packages/common/date/src/date.ts
+++ b/packages/common/date/src/date.ts
@@ -8,10 +8,10 @@ export type DateFnsDateType = number | Date;
 
 export const kstFormat = (date: DateFnsDateType, format: string) => _format(date, format, { locale });
 
-export const roundUpHoursInDays = (hours: number) => {
+export function roundUpHoursInDays(hours: number) {
   const remainder = hours % 24;
   return remainder === 0 ? hours : hours + 24 - (hours % 24);
-};
+}
 
 export function parseYYYYMMDD(yyyyMMdd: string) {
   const date = new Date(yyyyMMdd);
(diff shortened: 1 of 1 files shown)


packages/common/date/src/date.ts
```diff
@@ -8,10 +8,10 @@ export type DateFnsDateType = number | Date;
 
 export const kstFormat = (date: DateFnsDateType, format: string) => _format(date, format, { locale });
 
-export const roundUpHoursInDays = (hours: number) => {
+export function roundUpHoursInDays(hours: number) {
```
minsoo-web, line 11: I don't think it's a big issue, but if unifying is the main thing in this PR, shouldn't `kstFormat` be changed as well?

packages/common/date/src/date.ts
```diff
@@ -8,10 +8,10 @@ export type DateFnsDateType = number | Date;
...
   const remainder = hours % 24;
   return remainder === 0 ? hours : hours + 24 - (hours % 24);
-};
+}
```
minsoo-web, line 14: `"semi": true,` is defined in prettierrc file.

packages/common/date/src/date.ts
SEOKKAMONI: I agree that this is not a big issue but Wouldn't convention be important at work? That's why I posted it

I think I missed that part

SEOKKAMONI:
I don't think it's that important, so I'll close it sorry..

PR #328, opened 2023-08-12
docs(@toss/react): add flex.tsx example

SEOKKAMONI:
## Overview
add flex.tsx example

## PR Checklist


okinawaa (commented):
Is it better to have a document about flex function and flex component together?

PR #302, opened 2023-07-24
fix(utils): Merge @tossteam/utils

HoseungJang:
## PR Checklist

- [ ] I read and included theses actions below

1. I have read the [Contributing Guide](https://github.com/toss/slash/blob/main/.github/CONTRIBUTING.md)
2. I have written documents and tests, if needed.

PR #336, opened 2023-08-31
docs(@toss/utils): add @tossdocs-ignore to batchRequestsOf

chogyejin:
## Overview
In slash docs, there is [unneeded page](https://slash.page/ko/libraries/common/utils/src/batchRequestsOf.ts.tossdocs) because of `batchRequestsOf.ts`

스크린샷 2023-09-01 02 41 32

It seems that `@tossdocs-ignore` was accidentally deleted while importing the `noop` in #302, then I add it again like any other `.ts` files.

## PR Checklist

### DIFF
--- a/packages/common/utils/src/batchRequestsOf.ts
+++ b/packages/common/utils/src/batchRequestsOf.ts
@@ -1,3 +1,4 @@
+/** @tossdocs-ignore */
 import { noop } from './noop';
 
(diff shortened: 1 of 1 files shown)


raon0211 (commented):
Thanks, @chogyejin !

raon0211 (approved):

PR #337, opened 2023-09-08
docs(@toss/utils): add `@tossdocs-ignore` to `is-rrn.ts`

minjongbaek:
## Overview

In slash docs, there is [unneeded page](https://slash.page/ko/libraries/common/validators/src/validators/is-rrn.ts.tossdocs) because of `is-rrn.ts`

image

Looks like `@tossdocs-ignore` was deleted in #312, so I added it back like any other `.ts` file.
## PR Checklist


raon0211 (approved):
Thanks, @minjongbaek !

PR #339, opened 2023-09-18
chore(@toss/react): refac useVisibilityEvent hook

Sangminnn:
## Overview

I think already exist useDocumentVisibilityChange hook in useImpressionRef and that's better than useVisibilityEvent.

because that's more simple and abstract about 'document.visibilityState' ( visibilityState just has 'hidden' or 'visible' )

don't check document.visibilityState in use-case, just check isVisible in parameter.

## PR Checklist


raon0211:
Thanks for your contribution! 

Unfortunately, this leads to a breaking change to users of the library. For example, every user of the library has to update to use the `useDocumentVisibilityChange` hook after updating the library.

We try not to make breaking changes as much as possible unless there is a fair reason to do it; and I think the benefits we gain from this change is minor. 

So we decide to close this pull request, but thanks for your contribution again!

PR #342, opened 2023-10-06
fix(react): useCombinedRefs docs

ssi02014:
## Overview

I improved the `useCombinedRefs` documentation and replaced the `noop` used inside `useRefEffect` with the `noop` from `@toss/utils`. 🙏

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useCombinedRefs.spec.tsx
+++ b/packages/react/react/src/hooks/useCombinedRefs.spec.tsx
@@ -8,7 +8,7 @@ describe('useCombinedRefs', () => {
     const refs: Array<HTMLDivElement | null> = [];
     const callbackRef = jest.fn();
 
-    function TestComponent() {
+    const TestComponent = () => {
       const ref1 = useRef<HTMLDivElement>(null);
       const ref2 = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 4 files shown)


packages/react/react/src/hooks/useCombinedRefs.spec.tsx
```diff
@@ -8,7 +8,7 @@ describe('useCombinedRefs', () => {
     const refs: Array<HTMLDivElement | null> = [];
     const callbackRef = jest.fn();
 
-    function TestComponent() {
+    const TestComponent = () => {
```
okinawaa, line 11: Is it any benefit to use arrow function?
I'm curious because I couldn't find it

okinawaa (approved):

packages/react/react/src/hooks/useCombinedRefs.spec.tsx
ssi02014: @okinawaa 👋
I've seen some other test code that uses an arrow function to define "TestComponent", so I switched to that for consistency.

However, I've also seen test code written as function declarations, so there doesn't seem to be a rule. 

It doesn't seem to have a big impact, but should I change it back?

okinawaa (approved):
thank you!

PR #344, opened 2023-10-08
fix(react): update file extension

Lee-Seungje:
### update code readability
+ 4143398 : In JavaScript, undefined is treated as false by default. Therefore, you can use the `&&` operator directly without the need to write `!== undefined`. This helps improve readability by eliminating unnecessary code.
+ d48496e : It's sufficient to consider only the case where `i + 1` is equal to `childrenLength` without worrying about the case when it's not. There's no need for code that returns `null`.

### update code consistency
+ 3d61ece : In this commit, all files for exports other than the modified one are named `index.ts`. However, these files were originally named `index.tsx`, so they have been changed to `index.ts`. There's no need to use the `.tsx` extension for exports that do not use `JSX` syntax.

## PR Checklist


packages/react/react/src/components/OpenGraph/OpenGraph.tsx
```diff
@@ -11,9 +11,9 @@ interface Props {
 export function OpenGraph({ container: Container = Fragment, title, description, imageUrl }: Props) {
   return (
     <Container>
-      {title !== undefined && <meta property="og:title" content={title} />}
-      {description !== undefined && <meta property="og:description" content={description} />}
-      {imageUrl !== undefined && <meta property="og:image" content={imageUrl} />}
+      {title && <meta property="og:title" content={title} />}
```
okinawaa, line 14: I wonder what kind of readability went up while correcting this part!😄

## Explicitness and Intention clarity
**Explicit Undefined Checks**: Explicitly checking for **undefined** conveys a very clear intention that we are intentionally allowing other falsy values (like an empty string or null), which can be crucial in some use cases.

## Allowing All Non-Undefined Values
**Permitting Falsy Values:** We might have scenarios where falsy values (like an empty string) are valid and should render a <meta> tag. Using an implicit falsy check could exclude these valid cases and potentially introduce unexpected behavior.

While I greatly appreciate your effort and the intent behind simplifying checks, considering the aforementioned points, **I’d lean towards preserving the current explicit check against undefined**. However, I’m more than open to further discussion and any additional insights you might bring into this!

It might releated with [this comment](https://github.com/toss/slash/pull/345#discussion_r1349818842)

packages/react/react/src/components/Separated/Separated.tsx
```diff
@@ -15,7 +15,7 @@ export function Separated({ children, with: separator }: Props) {
       {childrenArray.map((child, i) => (
         <Fragment key={i}>
           {child}
-          {i + 1 !== childrenLength ? separator : null}
```
okinawaa, line 18: ensures the separator is rendered between all children except after the last one. Essentially, for every child except the last, the separator will be displayed.

Contrastingly, the code proposed in your PR:
**would render the separator only after the last child**, altering the current visual representation of our UI. This change could have a varied impact on all instances where this code pattern is used, potentially disrupting the established and intended UI pattern across our application.

I also had a hard time understanding this part.
I think it would be better to write a test code and clearly understand the intention of the existing code.

okinawaa (changes requested):
thank you 😄

packages/react/react/src/components/OpenGraph/OpenGraph.tsx
Lee-Seungje: Thank you for your feedback. I understand your intention well! This part has been restored in the following commit.
[91bc089](https://github.com/toss/slash/pull/344/commits/91bc08979ef3130d8b10dcbc9ae7af97859f4a34)

packages/react/react/src/components/Separated/Separated.tsx
Lee-Seungje: Thank you for your feedback. I didn't fully grasp the issue, and I will practice further review on the aspect I had in mind. This part has been restored in the following commit. Thank you.
[91bc089](https://github.com/toss/slash/pull/344/commits/91bc08979ef3130d8b10dcbc9ae7af97859f4a34)

okinawaa (approved):
Thank you!

PR #350, opened 2023-10-09
test(@toss/react): add convertNewlineToJSX test code

SEOKKAMONI:
## Overview

- [`03e62e9`](https://github.com/SEOKKAMONI/slash/commit/03e62e99cda0d713ceaa44c010df678f523cd4f1) add convertNewlineToJSX test code

image

## PR Checklist


packages/react/react/src/utils/convertNewlineToJSX.test.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { create } from 'react-test-renderer';
+import { convertNewlineToJSX } from './convertNewlineToJSX';
+
+describe('convertNewlineToJSX', () => {
+  it('converts newline characters into <br /> JSX elements', () => {
```
okinawaa, line 6: When creating a test case

I think it would be good to write 0, 1, n + 1 case.

Currently, it seems that only the situation with 1` \n` has been verified.
Can you write test cases when there is 0 `\n` test case at all or more than one `\n`?

```tsx

// zero \n
const input = 'There was a problem trying to sign in. please try again.';

// n + 1 \n
const input = 'There was a problem \ntrying to\n sign in. please\ntry again.';

```

packages/react/react/src/utils/convertNewlineToJSX.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <React.Fragment key={0}>There was a problem trying to sign in.</React.Fragment>,
+      <React.Fragment key={1}>
+        <br />
+        please try again.
+      </React.Fragment>,
+    ];
+
+    const result = convertNewlineToJSX(input);
+
+    result.forEach((element, index) => {
```
okinawaa, line 18: **I think the actual implementation of this function code is revealed in test code.**

I don't think it's good because the test is broken if the implementation changes. 
The purpose of this function is to add just `<br/>`, and `Fragment` doesn't seem to be interested in aspect of this function .

below paragraph is more natural I think!

```tsx
    const input = 'There was a problem trying to sign in.\nplease try again.';

    const result = convertNewlineToJSX(input);

    render(<>{result}</>);

    screen.debug();

    expect(screen.getByText('There was a problem trying to sign in.<br/>please try again.')).toBeInTheDocument();
```

packages/react/react/src/utils/convertNewlineToJSX.test.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { create } from 'react-test-renderer';
+import { convertNewlineToJSX } from './convertNewlineToJSX';
+
+describe('convertNewlineToJSX', () => {
+  it('converts newline characters into <br /> JSX elements', () => {
+    const input = 'There was a problem trying to sign in.\nplease try again.';
+    const output = [
+      <React.Fragment key={0}>There was a problem trying to sign in.</React.Fragment>,
```
okinawaa, line 9: If Fragment is replaced with <></> in the `convertNewlineToJSX` code, this test will be broken, is that okay?

okinawaa (changes requested):

packages/react/react/src/utils/convertNewlineToJSX.tsx
```diff
@@ -7,7 +7,7 @@ export function convertNewlineToJSX(str: string) {
 
   return chunks.map((line, index) => (
     <Fragment key={index}>
-      {index > 0 ? <br /> : ''}
+      {index > 0 && <br />}
```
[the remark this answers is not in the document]
SEOKKAMONI: I thought it would be better to write the code more cleanly using the && operator, rather than including unnecessary ''. how about like this?

packages/react/react/src/utils/convertNewlineToJSX.test.tsx
SEOKKAMONI: I didn't think about it Thank you so much!

packages/react/react/src/utils/convertNewlineToJSX.test.tsx
SEOKKAMONI: It's my mistake.

packages/react/react/src/utils/convertNewlineToJSX.tsx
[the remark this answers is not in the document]
SEOKKAMONI: @okinawaa I changed it to return null instead of an empty string. I think this way readability improves, and it helps avoid returning unnecessary things. Could you check if it [works](https://github.com/toss/slash/pull/350/commits/14ff32a3d3657f1b9062a33aeae3c2d767c0e303)?

SEOKKAMONI:
The commit history is too dirty, so I created a new pr

PR #351, opened 2023-10-12
refac(@toss/utils): Improved object utils function types 

ssi02014:
## Overview
I think `"keyof T"` for the "key" of an object and `"T[keyof T]"` for the type of the "value" is sufficient.

```ts
const obj = {
  a: 1,
  b: 2,
  c: 3,
} as const;

type keys = keyof typeof obj; // "a" | "b" | "c"
type values = (typeof obj)[keyof typeof obj]; // 1 | 2 | 3
```

So in the `Record<PropertyKey, unknown>` part, it would be clearer and better to write `T[keyof T]` for `unknown`.

```ts
// as-is
<T extends Record<PropertyKey, unknown>>

// to-be
<T extends Record<PropertyKey, T[keyof T]>>
```

The existing `ObjectKeys` seems to make the type complex.
we don't need to make `type assertions` after modifying a type.

```ts
// as-is
export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`;

export function objectKeys<Type extends Record<PropertyKey, unknown>>(obj: Type): Array<ObjectKeys<Type>> {
  return Object.keys(obj) as Array<ObjectKeys<Type>>;
}

// to-be 1
export function objectKeys<T extends Record<PropertyKey, T[keyof T]>>(obj: T): Array<keyof T> {
  return Object.keys(obj); // (*)
}
```

<br />

Of course, it seems possible to define and utilize the types `ObjectValues` and `ObjectKeys` anew.

```ts
// to-be 2
type ObjectValues<T extends Record<PropertyKey, T[keyof T]>> = T[keyof T];
type ObjectKeys<T extends Record<PropertyKey, T[keyof T]>> = keyof T;

export const objectKeys = <T extends Record<PropertyKey, ObjectValues<T>>>(
  obj: T
): Array<ObjectKeys<T>> => {
  return Object.keys(obj);
};
```
The type inference result is the same after the change!! 👍

### objectKeys
keys

<br />

### objectValues
values

<br />

### objectEntries
스크린샷 2023-10-12 오후 3 13 35

<br />

### omit
omit
<br />

## PR Checklist

### DIFF
--- a/packages/common/utils/src/object/object-keys.ts
+++ b/packages/common/utils/src/object/object-keys.ts
@@ -1,6 +1,4 @@
 /** @tossdocs-ignore */
-export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`;
-
-export function objectKeys<Type extends Record<PropertyKey, unknown>>(obj: Type): Array<ObjectKeys<Type>> {
-  return Object.keys(obj) as Array<ObjectKeys<Type>>;
+export function objectKeys<T extends Record<PropertyKey, T[keyof T]>>(obj: T): Array<keyof T> {
+  return Object.keys(obj);
 }
--- a/packages/common/utils/src/object/omit.ts
+++ b/packages/common/utils/src/object/omit.ts
@@ -1,12 +1,12 @@
 /** @tossdocs-ignore */
-import { ObjectKeys, objectKeys } from './object-keys';
+import { objectKeys } from './object-keys';
 import { ElementType } from './types';
 
-export function omit<ObjectType extends Record<PropertyKey, any>, KeyTypes extends Array<ObjectKeys<ObjectType>>>(
-  obj: ObjectType,
-  keys: KeyTypes
-) {
+export function omit<T extends Record<PropertyKey, T[keyof T]>, K extends Array<keyof T>>(
+  obj: T,
+  keys: K
+): Omit<T, ElementType<K>> {
   return objectKeys(obj)
-    .filter((k): k is Exclude<ObjectKeys<ObjectType>, ElementType<KeyTypes>> => !keys.includes(k))
-    .reduce((acc, key) => ((acc[key] = obj[key]), acc), {} as Omit<ObjectType, ElementType<KeyTypes>>);
+    .filter((k): k is Exclude<keyof T, ElementType<K>> => !keys.includes(k))
+    .reduce((acc, key) => ((acc[key] = obj[key]), acc), {} as Omit<T, ElementType<K>>);
 }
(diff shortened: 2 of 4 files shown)


packages/common/utils/src/object/object-keys.ts
```diff
@@ -1,6 +1,4 @@
 /** @tossdocs-ignore */
-export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`;
-
-export function objectKeys<Type extends Record<PropertyKey, unknown>>(obj: Type): Array<ObjectKeys<Type>> {
-  return Object.keys(obj) as Array<ObjectKeys<Type>>;
+export function objectKeys<T extends Record<PropertyKey, T[keyof T]>>(obj: T): Array<keyof T> {
+  return Object.keys(obj);
```
ssi02014, line 3: I guess we could also modify and recycle existing ObjectKeys as shown below. 

```ts
export type ObjectKeys<T> = keyof T;
```

Of course, we could also define new ObjectValues.

```ts
export type ObjectValues<T> = T[keyof T];
```

packages/common/utils/src/object/omit.ts
```diff
@@ -1,12 +1,12 @@
 /** @tossdocs-ignore */
-import { ObjectKeys, objectKeys } from './object-keys';
+import { objectKeys } from './object-keys';
 import { ElementType } from './types';
 
-export function omit<ObjectType extends Record<PropertyKey, any>, KeyTypes extends Array<ObjectKeys<ObjectType>>>(
-  obj: ObjectType,
-  keys: KeyTypes
-) {
+export function omit<T extends Record<PropertyKey, T[keyof T]>, K extends Array<keyof T>>(
```
HoseungJang, line 5: 1. The original code means "You should pass an object value to `obj`." so `unknown` seems appropriate. It doesn't have to be `T[keyof T]`.
2. The reason why `ObjectKeys` is used here is that `Object.keys` excludes symbol property, because symbol is not enumerable.

ssi02014: @HoseungJang 
Oh, Thank you..! The second part is something I missed.

```ts
export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<keyof T, symbol>}`;
```
One more question, is there any particular reason why you wrapped `Exclude` in backticks in the above sentence?
Do you think it would be okay to remove the backticks?

HoseungJang: The answer is no because `Object.keys` basically returns an array of strings. The backticks are the syntax of [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) and it is for converting `T` to string type.

ssi02014: @HoseungJang 
I understand what you're saying. 

I initially thought it would be okay to remove the vectic because the type inference would be the same, but I understand your intent. 

Thanks.

PR #355, opened 2023-10-29
fix(@toss/react): Fix the useOutsideClickEffect callback function to be called only once

### LINKED ISSUE #354: [BUG]: useOutsideClickEffect callback function runs twice in touch event.
### **Package Scope**

Package name: @toss/react/hooks/useOutsideClickEffect.ts

### **Describe the bug**

In a touch environment , the callback function runs twice. This can create unexpected bugs.

### **Expected behavior**

The callback function must be called only once in any environment.

### **To Reproduce**

https://codesandbox.io/s/thirsty-torvalds-58vq9q?file=/src/App.js

**use Chrom Mobile debugger or Phone**

스크린샷 2023-10-29 오후 6 06 46

### **Possible Solution**

There are many approaches to solving the problem.

The simplest (but not perfect) solution is to handle touch events only in a mobile environment and click events if not in a mobile environment.

I guessed that the main container added touchstart event handler because of delay issue of click event in mobile environment (or touchable environment)

sa02045:
## Overview

FIX #354 

Because of the [limitations on the touch event test in Testing library](https://github.com/testing-library/user-event/issues/880), no way to test was found. 
Let me know if there's any other way

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useOutsideClickEffect.ts
+++ b/packages/react/react/src/hooks/useOutsideClickEffect.ts
@@ -1,4 +1,4 @@
-import { isNotNil } from '@toss/utils';
+import { isMobileWeb, isNotNil } from '@toss/utils';
 import { useCallback, useEffect, useRef } from 'react';
 
 type OneOrMore<T> = T | T[];
@@ -31,12 +31,18 @@ export function useOutsideClickEffect(container: OneOrMore<HTMLElement | null>, 
   );
 
   useEffect(() => {
-    document.addEventListener('click', handleDocumentClick);
-    document.addEventListener('touchstart', handleDocumentClick);
+    if (isMobileWeb()) {
+      document.addEventListener('touchstart', handleDocumentClick);
+    } else {
+      document.addEventListener('click', handleDocumentClick);
+    }
 
     return () => {
-      document.removeEventListener('click', handleDocumentClick);
-      document.removeEventListener('touchstart', handleDocumentClick);
+      if (isMobileWeb()) {
+        document.removeEventListener('touchstart', handleDocumentClick);
+      } else {
+        document.removeEventListener('click', handleDocumentClick);
+      }
     };
   });
 }


okinawaa (commented):
Thank you for contributing.

I think you solved it well.

Since there is a problem with the touch device, can't we solve problem either device is touch device or not ?

`IsMobileWeb` is only branching according to `iOS`, `android`, not processing whether it's a touch device or not.

I think you can go over things like navigator.maxTouchPoints

minsoo-web:
Can't we listen to both `touch` and `mouse` events? 
Since mouse events don't run on touch devices, adding listeners won't cause any performance inconvenience.

sa02045:
> Can't we listen to both `touch` and `mouse` events? Since mouse events don't run on touch devices, adding listeners won't cause any performance inconvenience.


Thank you for your response.
In fact, user touches the screen both touch and click events will occur [reference](https://web.dev/articles/mobile-touchandmouse#1_-_clicking_and_tapping_-_the_natural_order_of_things)

Even when I experimented with my iPad, mouse events such as musedown occurred when I touched it

https://codesandbox.io/s/trusting-fog-6vgygn?file=/src/App.js

https://github.com/toss/slash/assets/50866506/ef17f5d3-d5b1-4daf-9ba4-8e427731fa47

sa02045:
> Thank you for contributing.
> 
> I think you solved it well.
> 
> Since there is a problem with the touch device, can't we solve problem either device is touch device or not ?
> 
> `IsMobileWeb` is only branching according to `iOS`, `android`, not processing whether it's a touch device or not.
> 
> I think you can go over things like navigator.maxTouchPoints

Yes. As you said, branching out depending on whether it's a touch device is one of the solutions.

But the problem with this approach is that there are devices like `touchable laptops.`

Desktop browsers support touch as well as mobile browsers. https://caniuse.com/?search=touch


The summary is as follows.

| Device              | Trigger Events(When touch or click) | Event to need to listen to                                                           |
| ------------------- | ----------------------------------- | ------------------------------------------------------------------------------------ |
| 📱 Phone(Pad)       | touch, click                        | ✅ touch <br/>  ❌ click (Now the problem is here!)                                                      |
| 💻 touchable laptop | touch, click                        | ✅ touch, ✅ click                                                                         |
| 💻 Laptop           | click                               | ✅ click <br/> ⚠️ touch (doesn't matter this listen to a touch event because this can't touch it) |




What if the code changes as follows?

```js
document.addEventListener('touchstart', handleDocumentClick);
if (!isMobileWeb()) {
  document.addEventListener('click', handleDocumentClick);
}
```

ssi02014:
@sa02045 
It's a interesting topic, and I'm keeping track of it!!  👍
Off topic, I think defining an eventType like this would reduce repetitive code 
```ts
useEffect(() => {
  const eventType = isMobileWeb() ? 'touchstart' : 'click';

  document.addEventListener(eventType, handleDocumentClick);
  return () => {
    document.removeEventListener(eventType, handleDocumentClick);
  };
});
```

HoseungJang:
The main problem we should solve is that the callback could be called twice, not which event must be triggered.

So we just make `useOutsideClickEffect` only listen click event to solve the problem. Click event is triggered regardless of whether a device has touchscreen or not.

```typescript
useEffect(() => {
  document.addEventListener('click', handleDocumentClick);

  return () => {
    document.removeEventListener('click', handleDocumentClick);
  };
});
```

sa02045:
> The main problem we should solve is that the callback could be called twice, not which event must be triggered.
> 
> So we just make `useOutsideClickEffect` only listen click event to solve the problem. Click event is triggered regardless of whether users interacted with screen by mouse or touch or something.
> 
> ```ts
> useEffect(() => {
>   document.addEventListener('click', handleDocumentClick);
> 
>   return () => {
>     document.removeEventListener('click', handleDocumentClick);
>   };
> });
> ```

Yes, listening to click events alone solves the problem easily.

However, listening only to click events can lead to poor usability for mobile users.
This is because there is a subtle difference in the trigger timing of click events and touchstart events.

A click event is triggered when a finger is removed from the touchscreen. However, in a mobile environment, users often swipe the screen or drag the screen without removing their finger.

This creates a "delay" in the action we want to invoke. Personally, I think this is a pretty inconvenient way to use it.

That's why the this function seems to listen to the touch start event just like any other [reference library](https://github.com/ndelvalle/v-click-outside/blob/c2e34581f89a750d08a377ed52a4c31cefff8aa1/src/v-click-outside.js#L10).

So we have to keep listening to the touch start event and work around it.

HoseungJang:
> However, listening only to click events can lead to poor usability for mobile users.
This is because there is a subtle difference in the trigger timing of click events and touchstart events.
>
> A click event is triggered when a finger is removed from the touchscreen. However, in a mobile environment, users often swipe the screen or drag the screen without removing their finger.
>
> This creates a "delay" in the action we want to invoke. Personally, I feel this has a lot of negative impact on usability.

But that's exactly why I told you that `useOutsideClickEffect` should listen only click events.

First of all, what is click? It is an action to press on and release an element with something like pointer, finger, ...

But how about the touchstart event? It is triggered when you just put your fingers on a screen. It is different from the click action and it makes `useOutsideClickEffect` work differently on non-touch and touch devices.

And you also told me about 'delay' and the swipe and drag gestures. But the gestures are also different from the click action, and they never trigger the click events. So I think it is not appropriate to call them 'delay'.

HoseungJang:
Simply put, I would like to ask you to think more about the name of `useOutside"Click"Effect`.

PR #267, opened 2023-06-16
test(@toss/date): correct unit test describe with 2019년 to 2020년

minsoo-web:
## Overview

Correct unit test describe with `2019` to `2020`

## PR Checklist

- [x] I read and included theses actions below

1. I have read the [Contributing Guide](https://github.com/toss/slash/blob/main/.github/CONTRIBUTING.md)
2. I have written documents and tests, if needed.

PR #359, opened 2023-11-15
fix(@toss/crypto): Removed duplicate code in powerMod function

po4tion:
## Overview

Recursive functions are one of the most complex ways to write code. So I made it more concise by reducing the duplication of the code. Code was also created to test the iterative recursive function.

If you don't need a test code, I'll remove it.

## PR Checklist

### DIFF
--- a/packages/common/crypto/src/pkcs1/utils.spec.ts
+++ b/packages/common/crypto/src/pkcs1/utils.spec.ts
@@ -2,28 +2,28 @@ import BigNumber from 'bn.js';
 import { powerMod } from './utils';
 
 describe('powerMod function tests', () => {
-  test('powerMod with small numbers', () => {
+  it('powerMod with small numbers', () => {
     const n = new BigNumber(2);
     const p = new BigNumber(3);
     const m = new BigNumber(5);
     expect(powerMod(n, p, m).toString()).toBe(new BigNumber(3).toString()); // 2^3 % 5 = 8 % 5 = 3
   });
 
-  test('powerMod with p = 1 should return n', () => {
+  it('powerMod with p = 1 should return n', () => {
     const n = new BigNumber(7);
     const p = new BigNumber(1);
     const m = new BigNumber(10);
     expect(powerMod(n, p, m).toString()).toBe(n.toString()); // 7^1 % 10 = 7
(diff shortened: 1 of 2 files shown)


packages/common/crypto/src/pkcs1/utils.spec.ts
```diff
@@ -0,0 +1,32 @@
+import BigNumber from 'bn.js';
+import { powerMod } from './utils';
+
+describe('powerMod function tests', () => {
+  test('powerMod with small numbers', () => {
```
okinawaa, line 5:
```suggestion
  it('powerMod with small numbers', () => {
```
you can refer https://github.com/toss/slash/pull/267#discussion_r1248051773

po4tion: Thank you for your advice. I modified the code as [follows]((https://github.com/toss/slash/pull/359/commits/56939f3cff36725eab0df1c9f77237e57ceedaec)).

packages/common/crypto/src/pkcs1/utils.ts
```diff
@@ -38,25 +38,22 @@ export function rsaep(n: BigNumber, e: BigNumber, m: BigNumber): BigNumber {
...
+    return squared;
   }
 
-  // const t = power_mod(n, p >> 1n, m);
-  const t = powerMod(n, p.shrn(1), m);
-
-  // return (t * t * n) % m;
-  return t.mul(t).mul(n).mod(m);
+  return squared.mul(n).mod(m); // (t * t * n) % m;
 }
```
okinawaa, line 58: There is an comment at the top of this file as below. 
If We keep the original code and sync, I think **it will be less confusing when maintaining it in the future**, so I don't think We should make any changes, what do you think?

>// 소스코드의 주석은 특별한 언급이 없다면 god_crypto의 원본 소스코드 입니다
// reference: https://github.com/invisal/god_crypto/blob/25ec83d77c73a5eec1f763fbc660a9442ce0b65e/src/rsa/rsa_internal.t#L73

po4tion: I agree with you.

PR #360, opened 2023-11-15
test(@toss/utils): Improve groupBy Function Tests (Translation and Case Sensitivity Enhancement)

JayeHa:
## Overview

1. Translate groupBy test descriptions from Korean to English for clarity and global team accessibility
2. Enhance groupBy tests with case sensitivity check
I added a new test case because the existing code's test cases were only in Korean, and there was no example case to check if the return value of createKey distinguishes between uppercase and lowercase when it's in English

## PR Checklist

### DIFF
--- a/packages/common/utils/src/groupBy.spec.ts
+++ b/packages/common/utils/src/groupBy.spec.ts
@@ -1,6 +1,6 @@
 import { groupBy } from './groupBy';
 
-describe('groupBy 함수는', () => {
+describe('The groupBy function', () => {
   const input = [
     { groupName: '부엉이', value: 1 },
     { groupName: '다람쥐', value: 2 },
@@ -10,7 +10,7 @@ describe('groupBy 함수는', () => {
     { groupName: '부엉이', value: 6 },
   ];
 
-  it('createKey의 return value를 이용해서 item을 묶는다', () => {
+  it('groups items using the return value of createKey', () => {
     expect(groupBy(input, ({ groupName }) => groupName)).toEqual({
       부엉이: [
         { groupName: '부엉이', value: 1 },
@@ -25,7 +25,7 @@ describe('groupBy 함수는', () => {
     });
   });
 
-  it('빈 Array를 넣었을 땐 빈 object를 return한다.', () => {
+  it('returns an empty object when given an empty array', () => {
     expect(groupBy([], ({ someKey }) => someKey)).toEqual({});
   });
 });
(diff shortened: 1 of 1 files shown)


packages/common/utils/src/groupBy.spec.ts
```diff
@@ -10,7 +10,7 @@ describe('groupBy 함수는', () => {
     { groupName: '부엉이', value: 6 },
   ];
 
-  it('createKey의 return value를 이용해서 item을 묶는다', () => {
+  it('groups items using the return value of createKey', () => {
```
okinawaa, line 13: I think `it groups` is wrong gramatically
```suggestion
  it('should group items using the return value of createKey', () => {
```

you can refer https://github.com/toss/slash/pull/267#discussion_r1248051773

packages/common/utils/src/groupBy.spec.ts
```diff
@@ -25,7 +25,25 @@ describe('groupBy 함수는', () => {
...
+      apple: [
+        { name: 'apple', value: 1 },
+        { name: 'apple', value: 2 },
+      ],
+      Apple: [{ name: 'Apple', value: 3 }],
+      banana: [{ name: 'banana', value: 4 }],
+    });
+  });
+
+  it('returns an empty object when given an empty array', () => {
```
okinawaa, line 46: same reason with above comment 😄

```suggestion
  it('should return an empty object when given an empty array', () => {
```

okinawaa (changes requested):

packages/common/utils/src/groupBy.spec.ts
JayeHa: Thank you for letting me know 👍

[chore: correct the test descriptions grammatically.](https://github.com/toss/slash/pull/360/commits/b6d7a8ce28e88d3194eef881977d3942911b3945)

packages/common/utils/src/groupBy.spec.ts
JayeHa: I modified it as you told me. thanks :)

[chore: correct the test descriptions grammatically.](https://github.com/toss/slash/pull/360/commits/b6d7a8ce28e88d3194eef881977d3942911b3945)

packages/common/utils/src/groupBy.spec.ts
```diff
@@ -25,7 +25,25 @@ describe('groupBy 함수는', () => {
     });
   });
 
-  it('빈 Array를 넣었을 땐 빈 object를 return한다.', () => {
+  it('should group items distinguishing between uppercase and lowercase', () => {
```
okinawaa, line 28: it is good

okinawaa (approved):

PR #361, opened 2023-11-16
docs(@toss/react): fix syntax error in example code

ctdlog:
## Overview

I identified a syntax error in the example code. Upon reviewing the test code, it appears that the example needs to be modified as follows:

```tsx
const [Provider, useContext] = buildContext<{ title: string }>('TestContext', null);

function Inner() {
  const context = useContext();

  return <h1>{context.title}</h1>;
}

function Page() {
  return (
    <Provider title="타이틀">
      <Inner />
    </Provider>
  );
}
```

## PR Checklist


okinawaa (approved):
thanks 😄

PR #363, opened 2023-11-16
fix(use-funnel): Fix NonEmptyArray type in model.ts.

jingeon27:
## Overview

NonEmptyArray type already exists in @toss/util. However, a readonly property has been added to the existing NonEmptyArray type with the same name in model.ts. Although they have the same name, they have additional properties, which can cause confusion.

- The NonEmptyArray type, which was referenced in many places, has been abstracted into StepType.
- StepType can perform the same type inference as before.

## PR Checklist

### DIFF
--- a/packages/react/use-funnel/src/models.ts
+++ b/packages/react/use-funnel/src/models.ts
@@ -1,2 +1,4 @@
+import { NonEmptyArray } from '@toss/utils';
+
 /** @tossdocs-ignore */
-export type NonEmptyArray<T> = readonly [T, ...T[]];
+export type StepsType = Readonly<NonEmptyArray<string>>;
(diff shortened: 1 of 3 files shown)


packages/react/use-funnel/src/models.ts
```diff
@@ -1,2 +1,4 @@
+import { NonEmptyArray } from '@toss/utils';
+
 /** @tossdocs-ignore */
-export type NonEmptyArray<T> = readonly [T, ...T[]];
+export type StepsType = Readonly<NonEmptyArray<string>>;
```
HoseungJang, line 4:
```typescript
export type ReadonlyNonEmptyArray<T> = Readonly<NonEmptyArray<T>>;
```

raon0211 (approved):

PR #370, opened 2023-11-18
Fix(@toss/hangul) : update test code and remove magic number

po4tion:
## Overview

[ What did i do ]
1. I switched the test code from Korean to English.
2. Changed from Loose Equivalence Comparison to Strict Equivalence Comparison. The reason is that "slash" is using "TypeScript" so you can clearly see the type you are returning.
3. I removed the magic number and added a constant.

## PR Checklist

### DIFF
--- a/packages/common/hangul/src/disassemble.ts
+++ b/packages/common/hangul/src/disassemble.ts
@@ -15,21 +15,21 @@ export function disassembleHangulToGroups(str: string) {
   for (const letter of str) {
     const disassembledComplete = disassembleCompleteHangulCharacter(letter);
 
-    if (disassembledComplete !== undefined) {
+    if (disassembledComplete != null) {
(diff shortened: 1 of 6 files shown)


packages/common/hangul/src/disassemble.ts
```diff
@@ -15,21 +15,21 @@ export function disassembleHangulToGroups(str: string) {
   for (const letter of str) {
     const disassembledComplete = disassembleCompleteHangulCharacter(letter);
 
-    if (disassembledComplete != null) {
```
okinawaa, line 18: `slash` prefer shallow comparable then strict 
because of flexbility

po4tion: Thank you for the information!

And i fixed [it](https://github.com/toss/slash/pull/370/commits/9aaf0d8782f038b7a06c35ebda584a84cc527937).

okinawaa (approved):
good👍

PR #371, opened 2023-11-20
fix(@toss/utils): ceilUnit parameter was missing

cha2hyun:
## Overview

Only floorToUnit was provided. 

## PR Checklist


packages/common/utils/src/Numbers.ts
```diff
@@ -34,14 +34,22 @@ export const floorToUnit = createNumberFormatterBy(Math.floor);
...
+  if (options.floorUnit > 0) {
+    val = floorToUnit(value, options.floorUnit || 1);
+  } else if (options.ceilUnit > 0) {
+    val = ceilToUnit(value, options.ceilUnit || 1);
+  }
 
-  if (flooredVal === 0) {
-    return '0';
+  if (val === 0) {
+    return "0";
```
po4tion, line 49: "slash" uses "singlequote".

You can refer this prettierrc file.

[ .prettierrc]
image

po4tion (changes requested):

packages/common/utils/src/Numbers.ts
cha2hyun: Thanks for reviewing. I changed it.

cha2hyun:
@raon0211 
Thank you for your review. 🚀

The document already had the ceilUnit parameter written on it, but it was not actually implemented. 
This "PR" corrected the part that did not work.

https://slash.page/libraries/common/utils/src/Numbers_formatToKRW.i18n
image

raon0211 (approved):
Thanks!

PR #388, opened 2023-12-09
feat(@toss/utils): Add delay test code and match parameter

po4tion:
## Overview

[ What did i do ]
1. Add delay test code.
2. Match dealy.ts and delay.\*.md's parameter name.

[ Test Coverage ]
- before
스크린샷 2023-12-09 오후 9 42 55

- after
스크린샷 2023-12-09 오후 9 17 06

## PR Checklist

### DIFF
--- a/packages/common/utils/src/delay.spec.ts
+++ b/packages/common/utils/src/delay.spec.ts
@@ -1,0 +1,20 @@
+import { delay } from './delay';
+
+describe('delay', () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('should return a promise which is resolved after given milliseconds', () => {
+    const milliseconds = 3000;
+    const promise = delay(milliseconds);
+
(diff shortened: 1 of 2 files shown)


packages/common/utils/src/delay.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('should return a promise which is resolved after given milliseconds', () => {
+    const milliseconds = 3000;
```
okinawaa, line 13:
```suggestion
    const DELAYING_TIME = 3000;
```

I hope the role of the variable is clear

po4tion: I agree with you. I think it's better to name a constant which means a role. I fix [it](https://github.com/toss/slash/pull/388/commits/37f1ccdd0cd47429cf2b93dc5120ec86fcd8c954).

okinawaa (approved):
Cool!

PR #389, opened 2023-12-09
feat(@toss/utils): Translate tests codes for `escapeHTML` and update docs to match parameter name

po4tion:
## Overview

[ What did i do ]
1. Match markdown's function parameter name and origin function parameter name.
2. Change korean to english in test code.

[ Proposal ]

The `default` case of the switch statement in "escapeHTML" can never be executed. That's because it's filtered out of "regex."

If default is not executed, it seems that the confusion of the code will only increase.
Can I change default to the 5th case?

You can refer [escapeHTML](https://github.com/toss/slash/blob/main/packages/common/utils/src/escapeHTML.ts).

## PR Checklist

### DIFF
--- a/packages/common/utils/src/escapeHTML.en.md
+++ b/packages/common/utils/src/escapeHTML.en.md
@@ -1,7 +1,7 @@
 # escapeHTML
 
 ```typescript
-function escapeHTML(str: string): string;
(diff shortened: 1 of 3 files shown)


raon0211 (approved):
Thanks!

PR #395, opened 2023-12-11
test(@toss/utils): Improve tests for get-set

po4tion:
## Overview

[ What did i do ]

1. Improve get-set's uncovered line.

[ Before ]
스크린샷 2023-12-11 오후 3 53 48

[ After ] 
스크린샷 2023-12-11 오후 3 52 32

## PR Checklist

### DIFF
--- a/packages/common/utils/src/get-set.spec.ts
+++ b/packages/common/utils/src/get-set.spec.ts
@@ -23,18 +23,6 @@ describe('get function', () => {
     expect(get(obj, 'a.c')).toBeNull();
     expect(get(obj, 'a.b.d')).toBeUndefined();
   });
-
-  it('should handle mixed array and object structures', () => {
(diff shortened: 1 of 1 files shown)


packages/common/utils/src/get-set.spec.ts
```diff
@@ -17,6 +17,12 @@ describe('get function', () => {
   it('should work correctly with empty objects', () => {
     expect(get({}, 'a')).toBeUndefined();
   });
+  it('should handle paths leading to undefined or null values inside nested objects', () => {
```
okinawaa, line 20: How about separate test cases which null and undefined?

I think it is better each test suite have one test case

po4tion: That's a good opinion. I separated the test cases into two describes.

okinawaa (approved):
Lgtm

PR #396, opened 2023-12-18
fix(use-funnel): use strict equal operator

Ubinquitous:
## Overview

The logic that previously used the default not equal operator to determine  
whether targetStep is null was replaced with the strict equal operator.

## PR Checklist


raon0211:
Actually, this is a coding convention in Toss to test against both `null` and `undefined`, but thanks for the contribution!

PR #397, opened 2023-12-18
test(use-overlay): Change korean to english in test code

Ubinquitous:
## Overview

- Change korean to english in close-reference.spec.tsx
- Change korean to english in options.spec.tsx

## PR Checklist

### DIFF
--- a/packages/react/use-funnel/src/Funnel.tsx
+++ b/packages/react/use-funnel/src/Funnel.tsx
@@ -18,7 +18,7 @@ export const Funnel = <Steps extends NonEmptyArray<string>>({ steps, step, child
 
   const targetStep = validChildren.find(child => child.props.name === step);
 
-  assert(targetStep !== null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
+  assert(targetStep != null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
 
   return <>{targetStep}</>;
 };
--- a/packages/react/use-funnel/src/Funnel.tsx
+++ b/packages/react/use-funnel/src/Funnel.tsx
@@ -18,7 +18,7 @@ export const Funnel = <Steps extends NonEmptyArray<string>>({ steps, step, child
 
   const targetStep = validChildren.find(child => child.props.name === step);
 
-  assert(targetStep != null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
+  assert(targetStep !== null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
 
   return <>{targetStep}</>;
 };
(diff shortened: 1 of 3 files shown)


packages/react/use-funnel/src/Funnel.tsx
```diff
@@ -18,7 +18,7 @@ export const Funnel = <Steps extends NonEmptyArray<string>>({ steps, step, child
 
   const targetStep = validChildren.find(child => child.props.name === step);
 
-  assert(targetStep != null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
+  assert(targetStep !== null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
```
raon0211, line 21: Actually this is intended; we use `!= null` deliberately in Toss, because we can check the equivalence for both `null` and `undefined`.

```suggestion
  assert(targetStep != null, `${step} 스텝 컴포넌트를 찾지 못했습니다.`);
```

packages/react/use-overlay/src/tests/options.spec.tsx
```diff
@@ -12,8 +12,8 @@ afterEach(() => {
   jest.useRealTimers();
 });
 
-describe('useOverlay는', () => {
-  describe('exitOnUnmount 옵션을', () => {
+describe('useOverlay is', () => {
```
raon0211, line 15:
```suggestion
describe('useOverlay', () => {
```

packages/react/use-overlay/src/tests/options.spec.tsx
```diff
@@ -50,7 +50,7 @@ describe('useOverlay는', () => {
       return <>{open && <OverlayCallerComponent exitOnUnmount={exitOnUnmount} />}</>;
     }
 
-    it('true로 설정하면 useOverlay를 호출한 컴포넌트가 unmount 될 때 같이 unmount 된다.', async () => {
+    it('If set to true, it will be unmounted when the component that called useOverlay is unmounted.', async () => {
```
raon0211, line 53:
```suggestion
    it('will be unmounted if exitOnUnmount is true and the component is unmounted.', async () => {
```

packages/react/use-overlay/src/tests/options.spec.tsx
```diff
@@ -68,7 +68,7 @@ describe('useOverlay는', () => {
       });
     });
 
-    it('false로 설정하면 useOverlay를 호출한 컴포넌트가 unmount 되어도 같이 unmount 되지 않는다.', async () => {
+    it('If set to false, it will not be unmounted even if the component that called useOverlay is unmounted.', async () => {
```
raon0211, line 71:
```suggestion
    it('will not be unmounted if exitOnUnmount is false and the component is unmounted.', async () => {
```

raon0211 (approved):

raon0211:
Thanks, @Ubinquitous !

PR #420, opened 2024-01-19
fix(@toss/validators): Increase the accuracy of verification

po4tion:
## Overview

[ What did i do ]
According to the Gregorian calendar, there are months when 30 or 31 days do not exist. For example, February 30 and April 31.
The current "isBirthDate6" is not responsible for this, even though it is a function to verify the correct date of birth. So I've supplemented this.

[ Problem ]
image

[ TO-BE ]
```ts
expect(isBirthDate6('950230')).toEqual(false);
expect(isBirthDate6('950431')).toEqual(false);
```
image

## PR Checklist

### DIFF
--- a/packages/common/validators/src/validators/is-birth-date-6.spec.ts
+++ b/packages/common/validators/src/validators/is-birth-date-6.spec.ts
@@ -11,6 +11,8 @@ describe('isBirthDate6', () => {
     expect(isBirthDate6('foobar')).toEqual(false);
     expect(isBirthDate6('000000')).toEqual(false);
     expect(isBirthDate6('960732')).toEqual(false);
+    expect(isBirthDate6('951301')).toEqual(false);
+    expect(isBirthDate6('950231')).toEqual(false);
     expect(isBirthDate6('')).toEqual(false);
   });
 });
--- a/packages/common/validators/src/validators/is-birth-date-6.ts
+++ b/packages/common/validators/src/validators/is-birth-date-6.ts
@@ -1,5 +1,27 @@
 /** @tossdocs-ignore */
 export function isBirthDate6(birthDate: string) {
-  const re = /^[0-9]{2}(?:0[1-9]|1[0-2])(?:0[1-9]|[1,2][0-9]|3[0,1])$/;
-  return re.test(birthDate);
+  if (!/^\d{6}$/.test(birthDate)) {
+    return false;
+  }
+
+  const year = parseInt(birthDate.substring(0, 2), 10);
+  const month = parseInt(birthDate.substring(2, 4), 10);
+  const day = parseInt(birthDate.substring(4, 6), 10);
+
+  if (month < 1 || month > 12) {
+    return false;
+  }
+
+  const daysInFebruary = isLeapYear(year) ? 29 : 28;
+  const daysInMonth = [31, daysInFebruary, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+
+  if (day < 1 || day > daysInMonth[month - 1]) {
+    return false;
+  }
+
+  return true;
+}
+
+function isLeapYear(year: number) {
+  return year !== 0 && year % 4 === 0;
 }


packages/common/validators/src/validators/is-birth-date-6.spec.ts
```diff
@@ -11,6 +11,9 @@ describe('isBirthDate6', () => {
     expect(isBirthDate6('foobar')).toEqual(false);
     expect(isBirthDate6('000000')).toEqual(false);
     expect(isBirthDate6('960732')).toEqual(false);
+    expect(isBirthDate6('951301')).toEqual(false);
+    expect(isBirthDate6('950230')).toEqual(false);
+    expect(isBirthDate6('950431')).toEqual(false);
```
okinawaa, line 16: I think there are too many expect phrases to verify on one test suite(`it`). 
Why don't you split it up in different test suite?

packages/common/validators/src/validators/is-birth-date-6.ts
```diff
@@ -1,5 +1,27 @@
...
+  if (day < 1 || day > daysInMonth[month - 1]) {
+    return false;
+  }
+
+  return true;
+}
+
+function isLeapYear(year: number) {
+  return year !== 0 && year % 4 === 0;
 }
```
ssi02014, line 27: @po4tion 👋 cc. @okinawaa 

The leap year check for the year 2000 is currently incorrect.  ex) `000229`
```ts
// Abnormal behavior 
[31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
isBirthDate6("000229"); // false

// Normal behavior 
[31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
isBirthDate6("000229"); // false
```
There are actually 29 days in February 2000.
https://ko.wikipedia.org/wiki/2000%EB%85%84_2%EC%9B%94

```suggestion
function isLeapYear(year: number) {
  return year % 4 === 0;
}
```
Wouldn't the above code be more accurate? 
Is there something I'm missing? 🤔

packages/common/validators/src/validators/is-birth-date-6.ts
```diff
@@ -1,5 +1,27 @@
...
+    return false;
+  }
+
+  const year = parseInt(birthDate.substring(0, 2), 10);
+  const month = parseInt(birthDate.substring(2, 4), 10);
+  const day = parseInt(birthDate.substring(4, 6), 10);
+
+  if (month < 1 || month > 12) {
+    return false;
+  }
```
ssi02014, line 13: A small suggestion, but why not define the numbers 1 and 12 as `JANUARY` and `DECEMBER` constants to manage them? 🤗
```suggestion
  if (month < JANUARY || month > DECEMBER) {
    return false;
  }
```

okinawaa: I like This comment awesome..!

minsoo-web: I love it too.
This code prevents the use of magic numbers

packages/common/validators/src/validators/is-birth-date-6.spec.ts
po4tion: @okinawaa I agree with you. [test: split test case](https://github.com/toss/slash/pull/420/commits/9af5499148cf0c6ec8d50327488d62a45b1e3505)

packages/common/validators/src/validators/is-birth-date-6.ts
po4tion: @ssi02014 Thank you for finding a bug! I apply the code you suggested. fix: catch 2000y bug

packages/common/validators/src/validators/is-birth-date-6.ts
po4tion: @ssi02014 Thanks! I agree with you.[refactor: change from magic number to constant variable](https://github.com/toss/slash/pull/420/commits/42f7101396dcf0e6e238c97652bd340b37a57f6f) (cc. @minsoo-web, @okinawaa )

raon0211:
Hello @po4tion, thanks for your contribution.

In my opinion, the best way to handle this is using the proper date library, since `Date`s are hard to handle properly. If we were to use `date-fns`, we might have used:

```tsx
import { parse, isValid } from 'date-fns';

const parsed = parse(date, "yyMMdd", new Date());
return isValid(parsed);
```

How about using the proper library instead of implementing the validation logic ourselves?

raon0211 (approved):
Thanks for your awesome work!

po4tion:
@raon0211 That's an incredible speed. Your addition was helpful, thank you!

PR #422, opened 2024-01-25
feat(@toss/utils): Add iterateEnum util function

ghdtjgus76:
## Overview
Add iterateEnum util function to `@toss/utils`

When using Object.keys method to iterate over the enum, the type inference is string.
```
const arr1 = Object.keys(Language); // string[]
```

With iterateEnum, accurate type inference to the enum type is possible without type assertion.
```
iterateEnum(Language); // Language[]
```

## PR Checklist

### DIFF
--- a/packages/common/utils/src/iterateEnum.en.md
+++ b/packages/common/utils/src/iterateEnum.en.md
@@ -1,0 +1,19 @@
+# iterateEnum
+
+Ensures accurate type inference when iterating over enum types.
+
+## Example
+
+```typescript
+enum Language {
+  KOR = 'KOR',
+  ENG = 'ENG',
+  JPN = 'JPN',
+}
+
+// When using Object.keys method to iterate over the enum, the type inference is string.
+const arr1 = Object.keys(Language); // string[]
+
+// With iterateEnum, accurate type inference to the enum type is possible without type assertion.
+iterateEnum(Language); // Language[]
+```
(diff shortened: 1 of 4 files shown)


ssi02014:
Actually, inside `iterateEnum`, you call `Object.values()`, but your description uses `Object.keys()` as an example, is there a reason why? 

An enum can have a different key and value.
```ts
enum Language {
  KOR1 = "KOR",
  ENG2 = "ENG",
  JPN3 = "JPN",
}

// or Numeric enums
enum Direction {
  Up = 1,
  Down = 2,
}

Object.keys(Direction); // [ '1', '2', 'Up', 'Down' ]
Object.values(Direction); // [ 'Up', 'Down', 1, 2 ]
iterateEnum(Direction); // [ 'Up', 'Down', 1, 2 ]
```
You can see that this result is not what you intended in your description and test code. 😭

### Related
[Numeric enums](https://www.typescriptlang.org/ko/docs/handbook/enums.html#%EC%88%AB%EC%9E%90-%EC%97%B4%EA%B1%B0%ED%98%95-numeric-enums)

ghdtjgus76:
I've updated the implementation and test code to consider reverse mapping for numeric enums.

ghdtjgus76:
To answer your question, I've added it to the description to compare the type inference when using the Object.keys() method with the Object.values() method. 

And, I've updated the implementation and test code to consider reverse mapping for numeric enums.

Thanks for your good comment.

raon0211:
Thanks for your contribution!

We would like to keep our library as compact as possible. Unfortunately, in our opinion, iterating over enums is not a usecase that occurs in JavaScript development.

Closing the pull request for this reason, but looking forward to your next contribution.

PR #423, opened 2024-01-26
feat(@toss/utils): Add `isNil` util function

SEOKKAMONI:
## Overview
Add `isNil` util function

## PR Checklist

### DIFF
--- a/packages/common/utils/src/isNil.en.md
+++ b/packages/common/utils/src/isNil.en.md
@@ -1,0 +1,15 @@
+# isNil
+
+Checks if the given value is `null` or `undefined`;
+
+```typescript
+function isNil<T>(value: T | null | undefined): value is T;
+```
+
+## Example
+
+```typescript
+isNil(null); // true
+isNil(undefined); // true
+isNil(1); // false
+```
--- a/packages/common/utils/src/isNil.en.md
+++ b/packages/common/utils/src/isNil.en.md
@@ -3,7 +3,7 @@
 Checks if the given value is `null` or `undefined`;
 
 ```typescript
-function isNil<T>(value: T | null | undefined): value is T;
+function isNil<T>(value: T | null | undefined): value is null | undefined;
 ```
 
 ## Example
--- a/packages/common/utils/src/isNil.ko.md
+++ b/packages/common/utils/src/isNil.ko.md
@@ -3,7 +3,7 @@
 주어진 값이 null이나 undefined인지 확인합니다.
 
 ```typescript
-function isNil<T>(value: T | null | undefined): value is T;
+function isNil<T>(value: T | null | undefined): value is null | undefined;
 ```
 
 ## Example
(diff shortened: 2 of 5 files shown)


packages/common/utils/src/isNil.ko.md
```diff
@@ -0,0 +1,15 @@
+# isNil
+
+주어진 값이 null이나 undefined인지 확인합니다.
```
SEOKKAMONI, line 3:
```suggestion
주어진 값이 null이나 undefined 인지 확인합니다.
```

packages/common/utils/src/isNil.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { isNil } from '.';
+
+describe('isNil should', () => {
```
SEOKKAMONI, line 3:
```suggestion
describe('isNil', () => {
```

ssi02014:
@SEOKKAMONI  👋
 I'm not a maintainer of slash, but here's a quick comment. 

Type narrowing is not working as intended.

```tsx
const test = (arg: any) => {
  if (isNil(arg)) {
    arg; // any, Should be type inferred as `null | undefined`.
  } else {
    arg; // never
  }
};
```

With improvements like the ones below, you'll be able to do type narrowing as intended. 

```ts
function isNil<T>(val: T | undefined | null): val is null | undefined {
  return val == null;
}
```
```ts
const test = (arg: any) => {
  if (isNil(arg)) {
    arg; // null | undefined
  } else {
    arg; // any
  }
};
```

packages/common/utils/src/isNil.ts
```diff
@@ -0,0 +1,4 @@
+/** @tossdocs-ignore */
+export function isNil<T>(val: T | undefined | null): val is T {
```
SEOKKAMONI, line 2:
```suggestion
export function isNil<T>(val: T | undefined | null): val is null | undefined {
```

packages/common/utils/src/isNil.en.md
```diff
@@ -0,0 +1,15 @@
+# isNil
+
+Checks if the given value is `null` or `undefined`;
+
+```typescript
+function isNil<T>(value: T | null | undefined): value is T;
```
ssi02014, line 6:
```suggestion
function isNil<T>(value: T | null | undefined): value is null | undefined;
```

packages/common/utils/src/isNil.ko.md
```diff
@@ -0,0 +1,15 @@
...
+
+```typescript
+function isNil<T>(value: T | null | undefined): value is T;
```
ssi02014, line 6:
```suggestion
function isNil<T>(value: T | null | undefined): value is null | undefined;
```

ssi02014 (commented):
looks good! 🤗
However, a final review by the maintainer of `slash` is required. 

The documentation file(.md) also looks like it needs to be fixed

okinawaa (approved):
thanks😊

Whether `isNil` will really increase the developer experience over `==` is a matter of consideration.

I think it's right to give the service developer the option to use.

PR #425, opened 2024-02-04
refactor(@toss/emotion-utils): SpacingProps: Replace `never` with `Omit` for children

SEOKKAMONI:
## Overview
SpacingProps: Replace `never` with `Omit` for children

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/spacing.en.md
+++ b/packages/react/emotion-utils/src/spacing.en.md
@@ -8,7 +8,6 @@ Components that give Space declaratively
 
 ```ts
 function Spacing(props: {
-  children?: never;
   // default: 'vertical'
   direction?: AxisDirection;
   size: CSSPixelValue;
--- a/packages/react/emotion-utils/src/spacing.ko.md
+++ b/packages/react/emotion-utils/src/spacing.ko.md
@@ -8,7 +8,6 @@ Space를 선언적으로 주는 컴포넌트
 
 ```ts
 function Spacing(props: {
-  children?: never;
   // default: 'vertical'
   direction?: AxisDirection;
   size: CSSPixelValue;
--- a/packages/react/emotion-utils/src/spacing.tsx
+++ b/packages/react/emotion-utils/src/spacing.tsx
@@ -4,13 +4,15 @@ import { memo } from 'react';
 import { coerceCssPixelValue } from './coerceCssPixelValue';
 import { AxisDirection, CSSPixelValue, ExtendHTMLProps } from './types';
 
-type SpacingProps = ExtendHTMLProps<
-  HTMLDivElement,
-  {
-    children?: never;
-    direction?: AxisDirection;
-    size: CSSPixelValue;
-  }
+type SpacingProps = Omit<
+  ExtendHTMLProps<
+    HTMLDivElement,
+    {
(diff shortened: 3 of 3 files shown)


okinawaa:
I want to know what effect this PR has 😊

SEOKKAMONI:
> I want to know what effect this PR has 😊

I should have written the explanation when I posted the PR. I'm sorry 

There are many reasons, but the most compelling one is that unnecessary automatic completion of "children" in props can create confusion for developers.

image

ssi02014:
I saw this pull request and was intrigued, so here's a small comment...! 🙏

```ts
type SpacingProps = Omit<React.HTMLProps<HTMLDivElement>, "size" | "children" | "direction"> & {
    children?: undefined;
    direction?: AxisDirection | undefined;
    size: CSSPixelValue;
}
```
The `SpacingProps` type has `children?: undefined` as shown above. 

```tsx
<Spacing direction="horizontal" size={40} />
```
Given the use case of Space, I don't think `children?: undefined` is a necessary type.
I agree with @SEOKKAMONI  that it's appropriate to exclude the `children` type.

SEOKKAMONI:
> I saw this pull request and was intrigued, so here's a small comment...! 🙏
> 
> ```ts
> type SpacingProps = Omit<React.HTMLProps<HTMLDivElement>, "size" | "children" | "direction"> & {
>     children?: undefined;
>     direction?: AxisDirection | undefined;
>     size: CSSPixelValue;
> }
> ```
> 
> The `SpacingProps` type has `children?: undefined` as shown above.
> 
> ```tsx
> <Spacing direction="horizontal" size={40} />
> ```
> 
> Given the use case of `Spacing`, I think it's appropriate to exclude the `children` type, the same as @SEOKKAMONI

I'm curious about why the type is explicitly set to 'undefined' for 'children' when it's not necessary. I think this change actually lowers the DX by providing unnecessary autocompletion to developers. 

and I'm also curious why size and direction are being excluded using Omit instead of ExtendHTMLProps.

I think I might be wrong understandIf I gave me wrong understanding understandingstanding, please tell me when I got wrong understand

Lastly, thank you for always provoking my curiosity! Your comments are a great help to my learning!!

ssi02014:
@SEOKKAMONI 👋
It seems like my intention wasn't properly conveyed! 😭
```tsx
type SpacingProps = ExtendHTMLProps<
  HTMLDivElement,
  {
    children?: never;
    direction?: AxisDirection;
    size: CSSPixelValue;
  }
>;
```
The type of the existing `SpacingProps` ultimately ends up being as follows!
```tsx
type SpacingProps = Omit<React.HTMLProps<HTMLDivElement>, "size" | "children" | "direction"> & {
    children?: undefined;
    direction?: AxisDirection | undefined;
    size: CSSPixelValue;
}
```

스크린샷 2024-02-11 오전 3 19 09

You can understand this by looking at the code where `ExtendHTMLProps` is defined...!
```tsx
export type ExtendHTMLProps<Elem extends HTMLElement, T> = Omit<HTMLProps<Elem>, keyof T> & T;
```
The key point is that I also agree with your opinion!! 🤗

raon0211 (approved):
Thanks!

PR #441, opened 2024-02-27
fix(@toss/utils): change inequality operator to strict inequallity operator

ho991217:
## Overview
This PR includes changes in 
- `@toss/utils/device/isAndroid.ts`
- `@toss/utils/device/isIOS.ts`
- `@toss/utils/device/isMacOS.ts`

Since `String.prototype.match()` function only returns an `Array`or `null`, there's no need of using [inequality operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Inequality) instead of [strict inequality operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality).
To prevent future type-related issues, using strict inequality operator should be recommended.

## PR Checklist

### DIFF
--- a/packages/common/utils/src/device/isAndroid.ts
+++ b/packages/common/utils/src/device/isAndroid.ts
@@ -5,5 +5,5 @@ export function isAndroid() {
     return false;
   }
 
-  return navigator.userAgent.match(/Android/i) != null;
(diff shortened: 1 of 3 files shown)


okinawaa (approved):
LGTM

Issue #428, opened 2024-02-07
[BUG]: createTypedLocalStorage type inference fails with type boolean


injoonH:
### **Package Scope**

Package name: `@toss/storage@1.3.1`

### **Describe the bug**

1. `createTypedLocalStorage<boolean>` returns `TypedStorage<false> | TypedStorage<true>`.
2. Its `set` method has type `(next: never) => void`.

### **Expected behavior**

1. `createTypedLocalStorage<boolean>` returns `TypedStorage<boolean>`.
2. Its `set` method has type `(next: boolean) => void`.

### **To Reproduce**

[TypeScript Playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoE8AOEAmBlMAeyjgHMIB5TMYQkAZwB5UA+ZAbwChkfl6iS5APwAuZAWJkIAbm69QwGnAA2ANRUBXCKLSyAvpxwQEyuFBQm49emiy4JgiMzZdeyTFGAA3OJGQBrCHRZNw9vXxR+SXIQ3gQ6figNBAEACkD0MUTQUgAaZEJqWgYdDGx8ASkqGgTnAEpYnnIwVLqxVGQAH2QQDWVlRr4IFpAIAA8wdrbkL0JgHEGTCDNWsVn5wbCfP3poYBVgAC8ZOR4tiOQjXc8D4-1OQ2NTc2RLa2QAOQ0AWwAjaDK9kq5GQ40gIBwNkBFWiTl6f2gLlOyFACHMVggqUIMBguzAOnh-yg03WC2RRjRy12WJxeIJPyJJLmZIMj0sLzAdmQACUIPQ+mBnMgALxoUETCAQmyE6DIISfBkAuwwxzIdrKhxSZyyCnPFAwDQgFLFV7oyDQgAyhAQKk15Gc6SCWTAnhAeQKRQSOgo30UzA1wMonoYznyAHIoo4wyxprz+cpBawdU8zPrDca6KbluaNXz6MU7U5WI7MnwXTl8oUaiUxD6-dDC9VikxWOHI1Jo7G+QLtQ94gwwGXYSKsxFLdbbYHGL9CIRlMsQCxUgAidvkZf5VzyECKfZqTQQMTwZS7Th6BoPNcQAB0eNSLq0F84nOwyAAgiO4z2Z3OFyxpEAA)

```ts
import { createTypedLocalStorage } from "@toss/storage/typed";

[...]

PR #442, opened 2024-02-28
fix(@toss/storage): Fix boolean type inference for TypedStorage 

saul-atomrigs:
## Overview

Fixes https://github.com/toss/slash/issues/428 and added test code.

Basically, I've changed the generic type `T` to `T extends boolean ? boolean : T`

## PR Checklist

### DIFF
--- a/packages/common/storage/src/typed/factory.ts
+++ b/packages/common/storage/src/typed/factory.ts
@@ -4,7 +4,10 @@ import { NumberTypedStorage, TypedStorage, TypedStorageOptions } from './storage
 
 type Result<T> = T extends number ? NumberTypedStorage : TypedStorage<T>;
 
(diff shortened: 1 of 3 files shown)


okinawaa (approved):
I think it is awesome solution

Basically, do you know why this problem occurred when it was boolean?

saul-atomrigs:
`false` is a type (as is `true`). They're [literal types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) that are both subtypes of `boolean`. From that link (you have to scroll down a bit):

> There’s one more kind of literal type: boolean literals. There are only two boolean literal types, and as you might guess, they are the types true and false. The type boolean itself is actually just an alias for the union true | false.

So in our case, by writing `T extends boolean ? boolean : T`, we can extend `true` or `false` to `boolean`.

Issue #352, opened 2023-10-20
[Question] why isServer should check `global`?


manudeli:
https://github.com/toss/slash/blob/cbfb2eb70799b8dfd25dcc440ca0924a6e7bb84f/packages/common/utils/src/device/isServer.ts#L4-L6

We reuse isServer many times in slash libraries, but I don't know why we should check `global` in isServer. could I get slash's intention why check not only `window` also `global` please?

### No global checking libraries
We can see no global checking in below links
- usehoos-ts(check only window): https://usehooks-ts.com/react-hook/use-isomorphic-layout-effect#hook
- react-use(check only window): https://github.com/streamich/react-use/blob/master/src/misc/util.ts#L21
- use-isomorphic-layout-effect(check only document): https://github.com/Andarist/use-isomorphic-layout-effect/blob/main/src/index.ts
- react-redux(check only window): https://github.com/reduxjs/react-redux/blob/8605088606403f6909597405c67db91d907db86e/src/components/connectAdvanced.js#L35-L41

PR #447, opened 2024-03-22
fix(@toss/utils): Add 'Deno' in globalThis for isServer()

saul-atomrigs:
## Overview

This PR makes two changes:
1. Removed `typeof global !== 'undefined'` according to [this issue](https://github.com/toss/slash/issues/352#issue-1953691398) and [this conversation](https://github.com/suspensive/react/pull/239#discussion_r1365379813)
2. Added `'Deno' in globalThis;`. [Tanstack/query](https://github.com/TanStack/query/blob/2ddf8036360279fcd6e3567859ec9c68fbfc024a/packages/query-core/src/utils.ts#L65) and [vercel/swr](https://github.com/vercel/swr/blob/cd22c46c4a0977e9d9afcd7b23f629e56e729344/src/_internal/utils/env.ts#L6) now use this checking for `isServer`. And we should use `globalThis` instead of `window` because it [became deprecated in Deno](https://github.com/denoland/deno/pull/22057).

## PR Checklist

### DIFF
--- a/packages/common/utils/src/device/isServer.ts
+++ b/packages/common/utils/src/device/isServer.ts
@@ -1,3 +1,5 @@
+/** @tossdocs-ignore */
+
 export function isServer() {
   return typeof window === 'undefined' || 'Deno' in globalThis;
 }
--- a/packages/common/utils/src/device/isServer.ts
+++ b/packages/common/utils/src/device/isServer.ts
@@ -1,6 +1,3 @@
-/** @tossdocs-ignore */
-declare const global: unknown;
-
 export function isServer() {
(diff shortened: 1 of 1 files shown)


packages/common/utils/src/device/isServer.ts
```diff
@@ -1,6 +1,3 @@
-/** @tossdocs-ignore */
-declare const global: unknown;
-
 export function isServer() {
```
raon0211, line 1:
```suggestion
/** @tossdocs-ignore */

export function isServer() {
```

raon0211 (approved):
This looks nice, thanks for your contribution! I wonder why we have `global` check in our `isServer()` call.

PR #450, opened 2024-03-22
refactor(@toss/react): change to using usePreservedCallback

ssi02014:
## Overview

We can use `usePreservedCallback` to keep a reference to the callback function and ensure that it is up to date. 

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/usePreservedCallback.ts
+++ b/packages/react/react/src/hooks/usePreservedCallback.ts
@@ -8,10 +8,7 @@ export function usePreservedCallback<Callback extends (...args: any[]) => any>(c
     callbackRef.current = callback;
   }, [callback]);
 
-  return useCallback(
-    (...args: any[]) => {
-      return callbackRef.current(...args);
-    },
-    [callbackRef]
-  ) as Callback;
+  return useCallback((...args: any[]) => {
+    return callbackRef.current(...args);
+  }, []) as Callback;
 }
--- a/packages/react/react/src/hooks/useTimeout.ts
+++ b/packages/react/react/src/hooks/useTimeout.ts
@@ -1,24 +1,13 @@
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
+import { usePreservedCallback } from './usePreservedCallback';
 
 /** @tossdocs-ignore */
 export function useTimeout(callback: () => void, delay = 0) {
-  const savedCallback = useRef<() => void>();
+  const savedCallback = usePreservedCallback(callback);
 
   useEffect(() => {
-    savedCallback.current = callback;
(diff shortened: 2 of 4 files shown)


packages/react/react/src/hooks/usePreservedCallback.ts
```diff
@@ -8,10 +8,7 @@ export function usePreservedCallback<Callback extends (...args: any[]) => any>(c
...
 
-  return useCallback(
-    (...args: any[]) => {
-      return callbackRef.current(...args);
-    },
-    [callbackRef]
-  ) as Callback;
+  return useCallback((...args: any[]) => {
+    return callbackRef.current(...args);
+  }, []) as Callback;
```
ssi02014, line 13: "ref" is meaningless, even when added to a dependency array.

packages/react/react/src/hooks/useTimeout.ts
```diff
@@ -1,24 +1,14 @@
-import { useEffect, useRef } from 'react';
+import { noop } from '@toss/utils';
+import { useEffect } from 'react';
+import { usePreservedCallback } from './usePreservedCallback';
 
 /** @tossdocs-ignore */
-export function useTimeout(callback: () => void, delay = 0) {
-  const savedCallback = useRef<() => void>();
+export function useTimeout(callback = noop, delay = 0) {
```
ssi02014, line 6: Other hooks like `useImpressionRef` already use `noop` as the default value for the callback function.

```tsx
export function useImpressionRef<Element extends HTMLElement>({
  onImpressionStart: _onImpressionStart = noop, // (*)
  onImpressionEnd: _onImpressionEnd = noop, // (*)
  // ...
}: Options)
```

Also, if we set the default value like this, we don't need to use conditional statements to determine whether the callback function exists or not.

I think it would make the code more readable.

packages/react/react/src/hooks/useTimeout.ts
```diff
@@ -1,24 +1,14 @@
-import { useEffect, useRef } from 'react';
+import { noop } from '@toss/utils';
+import { useEffect } from 'react';
+import { usePreservedCallback } from './usePreservedCallback';
 
 /** @tossdocs-ignore */
 export function useTimeout(callback: () => void, delay = 0) {
-  const savedCallback = useRef<() => void>();
+  const savedCallback = usePreservedCallback(callback ?? noop);
```
ssi02014, line 7: if we set the default value like this, we don't need to use conditional statements to determine whether the callback function exists or not.

```tsx
function handleTimeout() {
  // (*)
  if (savedCallback.current) {
    savedCallback.current();
  }
}
```

I think it would make the code more readable!! 💪

Issue #449, opened 2024-03-22
[Question]:useOutsideClickEffect dependancies


tjrdnjs1534:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: packages/react/react/src/hooks/useOutsideClickEffect.ts

### **Overview**

when useOutsideClickEffect, it will rerender everytime.
there is no dependacies on useOutsideClickEffect hook

### **Describe the solution you'd like**

when i use useOutsideClickEffect Hook, it should add eventListner once. and should remove it when component unmont.
i think 
```
  useEffect(() => {
    document.addEventListener('click', handleDocumentClick);
    document.addEventListener('touchstart', handleDocumentClick);

    return () => {
      document.removeEventListener('click', handleDocumentClick);
      document.removeEventListener('touchstart', handleDocumentClick);
    };
  });
```
to
```

  useEffect(() => {
    document.addEventListener('click', handleDocumentClick);
    document.addEventListener('touchstart', handleDocumentClick);

    return () => {
      document.removeEventListener('click', handleDocumentClick);
      document.removeEventListener('touchstart', handleDocumentClick);
    };
  }, []);
```
is there any reason there are no dependancies on this hook? 
im curious about it

ssi02014:
I think that's a good catch!
However, `handleDocumentClick` should be added to the dependency array of the useEffect you fixed.
  - Why? `handleDocumentClick` function is memoized because it is written as a useCallback. 

Alternatively, I think you could use the `usePreservedCallback` custom hook! 
  - I've done something like the PR below, but it hasn't been reviewed yet, so take it as a guide.
  - https://github.com/toss/slash/pull/450

```tsx
export function useOutsideClickEffect(container: OneOrMore<HTMLElement | null>, callback: () => void) {
  const containers = useRef<HTMLElement[]>([]);

  useEffect(() => {
    containers.current = (Array.isArray(container) ? container : [container]).filter(isNotNil);
  }, [container]);

  const handleDocumentClick = usePreservedCallback(({ target }: MouseEvent | TouchEvent) => {
    if (target === null) {
      return;
    }

    if (containers.current.length === 0) {
      return;
    }

    if (containers.current.some(x => x.contains(target as Node))) {
      return;
    }

    callback();
  });

  useEffect(() => {
    document.addEventListener('click', handleDocumentClick);
    document.addEventListener('touchstart', handleDocumentClick);

    return () => {
      document.removeEventListener('click', handleDocumentClick);
[...]

PR #451, opened 2024-03-24
fix(@toss/react): useOutsideClickEffect is called twice on touch event

saul-atomrigs:
## Overview

Fixes https://github.com/toss/slash/issues/354 and https://github.com/toss/slash/issues/449

I agree with [this answer](https://github.com/toss/slash/pull/355#issuecomment-1873193681). `click` event works fine on touchable screen too, so there is no need to have a `touch` event listener.

I believe another solution might be adding `e.preventDefault()` for the `touch` event. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events):
> ... [event.preventDefault()](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) to keep the browser from continuing to process the touch event (this also **prevents a mouse event from _also_ being delivered**).

However, I think this is a bit more over-engineering... so I decided to go with the first solution.

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useOutsideClickEffect.ts
+++ b/packages/react/react/src/hooks/useOutsideClickEffect.ts
@@ -32,11 +32,9 @@ export function useOutsideClickEffect(container: OneOrMore<HTMLElement | null>, 
 
   useEffect(() => {
     document.addEventListener('click', handleDocumentClick);
-    document.addEventListener('touchstart', handleDocumentClick);
 
     return () => {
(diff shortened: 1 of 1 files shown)


ssi02014:
@saul-atomrigs Hello! 👋
There is an already discussed pull request for this issue 🙏
https://github.com/toss/slash/pull/355

raon0211 (approved):
Thanks!

PR #452, opened 2024-03-26
feat(SwitchCase): Allow enum type for 'value' prop

jvn4dev:
## Overview

This PR enhances the SwitchCase component by allowing it to accept enum type for the value prop. This change provides more flexibility in using the SwitchCase component. The PR also includes updates to the test cases and documentation to reflect this new feature.

## PR Checklist

### DIFF
--- a/packages/react/react/src/components/SwitchCase/SwitchCase.en.md
+++ b/packages/react/react/src/components/SwitchCase/SwitchCase.en.md
@@ -1,10 +1,10 @@
 # SwitchCase
 
-Components that can use the switch-case syntax declaratively
+This is a component that allows you to use the switch-case syntax declaratively. The Case type can accept string, number, and enum.
 
 ## Example
 
-```jsx
+```tsx
 <SwitchCase
   value={status}
   // Depending on whether the status value is `'a'`, `'b'`, or `'c'`, the components below will be rendered.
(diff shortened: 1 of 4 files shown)


po4tion:
@jvn4dev 
Is it necessary to add the enum type?
According to this document, enum is not trishaded by bundlers such as rollup, so it is included in the bundle even if it is not actually used.
What's the point of using it even at the risk of these problems?

jvn4dev:
> Hello, @jvn4dev ! Is it necessary to add the enum type? According to this [document](https://engineering.linecorp.com/ko/blog/typescript-enum-tree-shaking), enum is not tree-shaking by bundlers such as rollup, so it is included in the bundle even if it is not actually used. What's the point of using it even at the risk of these problems?

I knew that TypeScript enum is not trichaking, but sometimes it's used in projects for code readability.
I use union type as much as possible, but it's a bit unfortunate that I'm forcing my open source Toss library not to use enum, so I created that PR. Even if enum isn't good, the type script still has enum :)
When I used enum in the project to check the value, I thought it would be worse to deal with that type through type assertion like 'as unknown as string'.
Of course, I fully agree that not using the enum itself is ultimately the best way to go! But projects don't always flow in the best direction :(

po4tion:
> Thx for your reply @po4tion ! I knew that TypeScript enum is not tree shaked by bundlers, but sometimes it's used in projects for code readability. I use union type as much as possible, but it's a bit unfortunate that I'm forcing my open source Toss library not to use enum, so I created that PR. Even if enum isn't good, the type script still has enum :) When I used enum in the project to check the value, I thought it would be worse to deal with that type through type assertion like 'as unknown as string'. Of course, I fully agree that not using the enum itself is ultimately the best way to go! But projects don't always flow in the best direction :(

I don't understand that I give up something bigger because of the readability of the code.
"enum" generates additional code at compilation time, which increases the size of the final file, which can negatively affect the app's loading speed and performance.
However, I know that open source should be as equal to all developers as possible. I don't think we need to stop developers using enum from trying.
I think your contribution is great. Thank you for sharing your precious opinion!

raon0211:
Thanks for your contribution! 

I understand that enums are natively supported by TypeScript. However, as @po4tion mentioned, I think supporting numbers and strings in `<SwitchCase />` is enough.

Closing the pull request for this reason, but looking forward to your next contribution.

PR #455, opened 2024-03-27
 fix(@toss/utils) added improvements and test code to the queryString function.

Gaic4o:
## Overview

Hello. While modifying the query String function, I organized it into docs, fixes, and test formats.

**docs**: Translated existing comments and test code within `queryString.spec.ts`, `queryString.ts` from Korean to English.

**Fix**: Refactored the test case to use the QS module instead of calling the function directly. These changes will ensure that test cases more closely match real-world usage patterns when used in production in the QS modular format.

**Test**: We improved coverage by adding test cases for QS.get.


raon0211 (approved):
Thanks for your contribution! Didn’t know that we still have Korean in our codebase.

PR #467, opened 2024-04-15
test(@toss/react): Test hook instead of component

sa02045:
## Overview

Test Hook instead of the components wrapped around it

It is more meaningful to test the hook directly than to test the components wrapped around it.

We can eliminate concerns about the component itself (button, parograph, textContent, click event, etc.)

## PR Checklist


raon0211:
Thanks for your contribution! 

Unfortunately this is our one of the code conventions that test codes should be as similar as possible with code that we actually write. 

In my opinion, since we do not use the hooks by themselves, testing them using test components is more similar with what we actually write.

Closing the pull request for this reason, but thanks for your contribution!

Issue #463, opened 2024-04-11
[Feature]: Use relational operators for date comparison


injoonH:
### **Package Scope**

- [x] Add to an existing package
- [ ] New package

Package name: `@toss/date@1.1.8`

### **Overview**

As JavaScript `Date` objects support [relational operations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#relational_operators), using libraries like `date-fns` is not necessary.

### **Describe the solution you'd like**

```ts
export function isEqualOrBefore(lhs: Date, rhs: Date) {
  return lhs <= rhs;
  // return isEqual(lhs, rhs) || isBefore(lhs, rhs);
}

export function isEqualOrAfter(lhs: Date, rhs: Date) {
  return lhs >= rhs;
  // return isEqual(lhs, rhs) || isAfter(lhs, rhs);
}
```

### **Additional context**

- [Stack Overflow: Compare two dates with JavaScript](https://stackoverflow.com/a/493018)

PR #471, opened 2024-04-28
chore(@toss/date): use relational operators to compare dates

gunhoflash:
## Overview

Use relational operators to compare dates.

From issue #463 

## PR Checklist


raon0211:
Thanks for your contribution!

While it is true that dates can be compared with relational operators, it is because dates are implicitly converted numbers before they are being compared. 

Personally, I prefer to use more explicit functions to compare dates; this is why we are using functions from date-fns. 

Closing the pull request for this reason. But we appreciate your pull request!

PR #473, opened 2024-05-05
chore: remove footer i18n on docusaurus-theme-classic

cruelladevil:
## Overview

 
copyright year is fixed because of i18n footer options.
remove these options so that copyright year would be current year.

## PR Checklist


raon0211 (approved):
Thanks!

PR #484, opened 2024-05-29
test(@toss/react): improve consistency in `OutsideClick` test code

Collection50:
## Overview

Improve improve consistency in `OutsideClick` test code.

### ASIS

```tsx
it('should expose internal DOM element through external ref', () => {
  const ref = createRef<HTMLDivElement>();
  const onEffect = jest.fn();

  render(
    <OutsideClick ref={ref} callback={onEffect}>
      <div>Inside Component</div>
    </OutsideClick>
  );

  const insideElement = screen.getByText('Inside Component');
  expect(ref.current?.contains(insideElement)).toBeTruthy();
});
```

<br/>

### TOBE

```tsx
it('should expose internal DOM element through external ref', () => {
  const ref = createRef<HTMLDivElement>();
  const onEffect = jest.fn();

  render(
    <OutsideClick ref={ref} callback={onEffect}>
      <div data-testid="inside">inside</div>
    </OutsideClick>
  );

  const insideElement = screen.getByTestId('inside');
  expect(ref.current?.contains(insideElement)).toBeTruthy();
});
```

Another test code is being tested based on getByTestId, so I corrected it.

## PR Checklist

### DIFF
--- a/packages/react/react/src/components/OutsideClick/OutsideClick.test.tsx
+++ b/packages/react/react/src/components/OutsideClick/OutsideClick.test.tsx
@@ -51,11 +51,11 @@ describe('OutsideClick', () => {
 
     render(
       <OutsideClick ref={ref} callback={onEffect}>
-        <div>Inside Component</div>
+        <div data-testid="inside">inside</div>
       </OutsideClick>
     );
 
-    const insideElement = screen.getByText('Inside Component');
+    const insideElement = screen.getByTestId('inside');
     expect(ref.current?.contains(insideElement)).toBeTruthy();
   });
 });


ssi02014:
@Collection50 Hi! 👋
I would like to share a small opinion. I believe that tests based on `getByText` are better than those based on `getByTestId`. Furthermore, the best practice is to use `getByRole` for testing.

How about writing test code according to the [priority guidelines](https://testing-library.com/docs/queries/about#priority) suggested by testing-library?

Collection50:
> @Collection50 Hi! 👋 I would like to share a small opinion. I believe that tests based on `getByText` are better than those based on `getByTestId`. Furthermore, the best practice is to use `getByRole` for testing.
> 
> * `getByTestId` is the last resort 😂
> 
> How about writing test code according to the [priority guidelines](https://testing-library.com/docs/queries/about#priority) suggested by testing-library?
> 
> ### Note
> * [ARIA Role - MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)

Thank you so much for reviewing my code!

The test method written in the [`most components`](https://github.com/toss/slash/blob/60b035bcb94fbb73394c6703bd2f20f53ed14916/packages/react/react/src/components/Separated/index.spec.tsx#L9) code is `data-testid`, so I applied it like this to maintain consistency!


Following the instructions in the official document(you attatched) is much better for maintenance of the code. 
So with the new [PR](https://github.com/toss/slash/pull/485), I've made it possible to unify the use of `getByRole` on all packages!

raon0211:
Thanks @Collection50 for your contribution!

As @ssi02014 mentioned, I believe selecting elements with `getByText()` is a better choice when it comes to writing test codes. Closing the pull request for this reason.

Collection50:
> Thanks @Collection50 for your contribution!
> 
> As @ssi02014 mentioned, I believe selecting elements with `getByText()` is a better choice when it comes to writing test codes. Closing the pull request for this reason.

Thank you for your review.

The intention of this PR is to standardize the test method used across all test cases.

Currently, only one out of three test cases uses `getByText`. I propose unifying all the test cases to use `getByRole` for consistency. What are your thoughts on this approach?

PR #485, opened 2024-05-29
test(@toss/react): adjust test priority guidlines in `react-testing-library`

Collection50:
## ASIS

```tsx
<div css={gutter(direction)}>
  <div data-testid="child" />
  <div data-testid="child" />
</div>
```

## TOBE

```tsx
<div css={gutter(direction)}>
  <div role="child" />
  <div role="child" />
</div>
```

<br/>

I changed the data-testid to role according to the priority suggested by [`testing-library` official Document](https://testing-library.com/docs/queries/about/#priority).

<br/>

image

<br/>

## PR Checklist

### DIFF
--- a/packages/react/emotion-utils/src/SafeArea.test.tsx
+++ b/packages/react/emotion-utils/src/SafeArea.test.tsx
@@ -16,7 +16,7 @@ describe('SafeArea', () => {
 
     it(`SafeArea는 as를 통해 as의 태그로 렌더링된다`, () => {
       const { container } = render(
-        <SafeArea as="a" href="asdf" data-testid="root">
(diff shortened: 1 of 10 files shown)


raon0211:
Thanks for your contribution! 

Actually, I believe the guidelines in the testing libraries is to select elements with the role if possible. In this case, since `<div />` s do not have an actual role when it comes to accessibility, I believe selecting them with test ids is the best choice. I think adding some random role would not be the library author's intention. (For your reference, [here are the predefined roles that we can use](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles).)

But we are open to new contributions, thanks for your great PR!

PR #487, opened 2024-05-31
fix(@toss/hangul): replace non-null assertion(`!`) with nullish coalescing operator (`??`)

saul-atomrigs:
## Overview

https://github.com/toss/slash/blob/60b035bcb94fbb73394c6703bd2f20f53ed14916/packages/common/hangul/src/disassembleCompleteHangulCharacter.ts#L30-L32

This PR addresses the typescript-eslint warning (forbidden non-null assertion) by replacing it with the nullish coalescing operator (`??`) and giving an empty string as the default value.

#### AS-IS
Captura de pantalla 2024-05-31 a las 10 14 07 a  m

#### TO-BE
Captura de pantalla 2024-05-31 a las 10 24 32 a  m

source: https://typescript-eslint.io/rules/no-non-null-assertion/

## PR Checklist


raon0211:
Thanks for your contribution! 

Actually this was intended on purpose; there is no chance that the array access results in returning undefined. Adding nullish coalescing will add extra runtime overhead and increase in bundle size.

If possible, I think adding some comments like `// eslint-disable-next-line` is a better design choice.

saul-atomrigs:
Oh, I see your point makes sense. Thanks!

Would you like me to add either `// eslint-disable-next-line` above each line or maybe `/* eslint-disable @typescript-eslint/no-non-null-assertion */` for the entire page? Happy to contribute :)

raon0211:
I think adding separate `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` for each line would be helpful for us :)

saul-atomrigs:
Changes applied. Many thanks!

raon0211 (approved):
Thanks!

PR #489, opened 2024-06-02
feat(@toss/react): add a new useUnmount hook

ssi02014:
## Overview

Hello!! 👋

I have created a Pull Request to propose the addition of a new hook called `useUnmount`. This custom hook calls the callback function passed as an argument when a component is unmounted. 

It is intended to be used declaratively for performing specific actions upon unmounting.

It seems that existing hooks provided by slash can also be improved through the use of the `useUnmount` hook.

```diff
export function useThrottle<F extends (...args: any[]) => any>() {
  // ...
-  useEffect(
-   () => () => {
-     throttledCallback.cancel();
-   },
-   [throttledCallback]
-  );
+ useUnmount(() => throttledCallback.cancel());

  return throttledCallback;
}
```
```diff
export function useDebounce<F extends (...args: any[]) => any>(
  callback: F,
  wait: number,
  options: Parameters<typeof debounce>[2] = {}
) {
  // ...
  
-  useEffect(() => {
-    return () => {
-      debounced.cancel();
-    };
-  }, [debounced]);
+  useUnmount(() => debounced.cancel());

  return debounced;
}
```
I believe the readability has improved after the modifications, as shown in the example above! What do you think? 🤗 
I expect this can be utilized in various ways beyond the example provided. 

I would appreciate your positive review. 🙏

## References
- [react-use: useUnmount](https://github.com/streamich/react-use/blob/master/docs/useUnmount.md)
- [usehooks-ts: useUnmount](https://usehooks-ts.com/react-hook/use-unmount)

## PR Checklist

### DIFF
--- a/packages/react/react/src/hooks/useUnmount.ts
+++ b/packages/react/react/src/hooks/useUnmount.ts
@@ -1,0 +1,9 @@
+import { useEffect } from 'react';
+import { usePreservedCallback } from './usePreservedCallback';
+
+/** @tossdocs-ignore */
+export function useUnmount(callback: () => any){
+  const preservedCallback = usePreservedCallback(callback);
+
+  useEffect(() => () => preservedCallback(), [preservedCallback]);
+};
--- a/packages/react/react/src/hooks/index.ts
+++ b/packages/react/react/src/hooks/index.ts
@@ -26,4 +26,6 @@ export * from './useThrottle';
 export * from './useTimeout';
 export * from './useTimeoutQueue';
 export * from './useToggleState';
+export * from './useUnmount';
 export * from './useVisibilityEvent';
+
--- a/packages/react/react/src/hooks/useUnmount.en.md
+++ b/packages/react/react/src/hooks/useUnmount.en.md
@@ -1,0 +1,15 @@
+# useUnmount
+
+A hook that calls a callback function passed as an argument when the component is `unmounted`.
+
+```ts
+function useUnmount(callback: () => void): void
+```
+
+## Example
+
+```ts
+useUnmount(() => {
+  console.log("unmount");
+});
+```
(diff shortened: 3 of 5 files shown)


packages/react/react/src/hooks/useUnmount.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect } from 'react';
+import { usePreservedCallback } from './usePreservedCallback';
+
+/** @tossdocs-ignore */
+export function useUnmount(callback: () => any){
+  const preservedCallback = usePreservedCallback(callback);
```
ssi02014, line 6: By using usePreservedCallback, we maintain the reference and ensure it is up-to-date.

minsoo-web:
I agree with @raon0211, the interface I used to have in my head had the following hooks. 

```ts
useLifeCycle.mount(callback)

useLifeCycle.updated(callback, [deps])

useLifeCycle.unmount(callback)
```

The inspiration for the above interface came from the fact that, as @ssi02014  mentioned
> It is intended to be used declaratively for performing specific actions upon unmounting.

I was looking forward to being able to control the lifecycle of react **declaratively**.

However, the hooks that handle lifecycle could feel **overly abstracted** if they were not managed in `useEffect`.

In this PR, we expect to be able to control lifecycle **declaratively** for unmount, but we're also concerned about the **over-abstraction** of the clear function's control statements, which are not managed together inside the `useEffect`.

ssi02014:
@minsoo-web 
Thank you for your feedback! 🙏
I understand the concerns about `effects` and `cleanup` functions not being managed together.

I appreciate that there was already a reference to `useUnmount`, and I created this Pull Request believing there was room for improvement in the existing code. 😂 Thank you!

ssi02014:
@raon0211  @minsoo-web 
https://react.dev/reference/react/useEffect#my-cleanup-logic-runs-even-though-my-component-didnt-unmount
See also the article above.

raon0211:
Looking forward to your next contribution, closing the pull request.

PR #492, opened 2024-06-03
fix(@toss/utils): replace 힇 to 힣

novice0840:
## Overview

When using regular expressions to determine Korean characters, we are using [가-힣]. However, in the maskExceptForEdge method, the library is using “힇”. '힣's unicode value is more higher than '힇'. So if you want to contain whole hangul, you should use '힣'

Fix the typo and send a PR. 

## Attach an example of how you implemented the bug. 
image

## PR Checklist


raon0211 (approved):
Nice catch. Thanks, @novice0840 !

